PR #2, opened 2020-05-12
⚠️ Creation of CRDs and controller

maelk:
**What this PR does / why we need it:**
This PR adds the repository content, including:

- documentation
- api definition
- CRDs and deployment files
- tools and hacks
- controller
- Makefile


maelk:
This is mostly a light-weight copy of CAPM3, with some changes in the api, ipam and controller folders

russellb:
unrelated to this PR, but since it's a new repo ...

I would consider renaming the repo.  It's easier when it's in the metal3-io namespace, but if someone forks it to whatever/ipam, I think the name may be too generic.  k8s-ipam?  metal3-ipam?  If you can't think of a good other name, then it's not a huge deal.  Just my first reaction when i saw it ...

maelk:
yes definitely @russellb ! I opened the question on our slack channel. metal3-ipam would work for me.

dhellmann:
This is a really big PR. I don't know if there's any way to start smaller and add changes more incrementally, since a lot of the initial code is generated. Maybe we should land it and iterate on it from there?

maelk:
I was also wondering how to break it down. If I take out the go code, in a dedicated PR, all generated files come along, so the size does not reduce that much. The rest is directly copied from CAPM3

maelk:
@dhellmann I am going to land this and do a v0.0.1 release to be able to use it properly in CAPM3. We will iterate from here.

PR #7, opened 2020-06-12
🐛 Trim trailing - in IPAddress object names

maelk:
This fixes a bug when the ip address ends with 0 in ipv6, leading to a
name ending with "--". Those are now trimed

### DIFF
--- a/ipam/ippool_manager.go
+++ b/ipam/ippool_manager.go
@@ -278,11 +278,8 @@ func (m *IPPoolManager) createAddress(ctx context.Context,
 	}
 
 	if allocatedAddress, ok := m.IPPool.Status.Allocations[addressClaim.Name]; ok {
-		formatedAddress := strings.Replace(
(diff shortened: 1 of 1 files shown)


ipam/ippool_manager.go
```diff
@@ -518,3 +509,10 @@ func addOffsetToIP(ip, endIP net.IP, offset int) (net.IP, error) {
 	copy(ip[16-IPBytesLen:], IPBytes)
 	return ip, nil
 }
+
+// formatAddressName renders the name of the IPAddress objects
+func (m *IPPoolManager) formatAddressName(address ipamv1.IPAddressStr) string {
```
furkatgofurov7, line 514: It looks okay for me, although I am not that familiar with base code in this repo, but `address` shouldn't be called here as `addresses`

maelk: it's only one at a time

furkatgofurov7: Makes sense

PR #9, opened 2020-07-17
✨Add dnsServer to IPPool and IPAddress

### LINKED ISSUE #8: add nameserver info in IPPool
**User Story**

As a operator, I would like to use specific nameservers for IPs allocated from an IPPool. 

**Detailed Description**

correct nameservers need to be set matching the IP/network to do dns resolution.

/kind feature

maelk:

This commit adds the dnsServer field to the IPPool object, both on a per
pool and global basis within that object. It also adds it in the
IPAddress object, populated based on the value in the IPPool

**Which issue(s) this PR fixes** :
Fixes #8


api/v1alpha1/ipaddress_types.go
```diff
@@ -45,6 +45,9 @@ type IPAddressSpec struct {
 
 	// Address contains the IP address
 	Address IPAddressStr `json:"address"`
+
+	// DNSServers is the list of dns servers
+	DNSServers []IPAddressStr `json:"dnsServers,omitempty"`
```
dhellmann, line 50: I'm surprised that it's possible to add new fields to the type here without also updating the generated CRD file.

maelk: that's my bad, I messed up the commit and that got left out. I'll update. That also shows that we need the generation job here.

jzhoucliqr, line 50: Thanks for adding this so quickly!  One question, shall we add search domain together with address?

maelk: hum, I am not sure whether we would have a use case with nodes in different search domains. Would you have one ?

maelk: https://docs.openstack.org/nova/latest/_downloads/9119ca7ac90aa2990e762c08baea3a36/network_data.json does not support search domain, so from the BMO point of view we do not have a use-case. However, if you have one, there would be no problem adding it to support it.

jzhoucliqr: Thanks. It wouldn't necessarily be in different search domains, I'm just looking for ways to CONFIGURE the search domain, feels like a good place to go together with the dnsserver address. Please let me know your thoughts.

maelk: In a way yes, however, the search domain is something that is more static across your hosts, i.e. all your hosts are probably configured the same way. In that case it is rather something that belongs to the template used for your hosts, rather than the per-node values that are rendered to fill the cloud-init templates. So the search domain configuration could be part of the Infrastructure machine template or bootstrap provider template, depending on your providers implementations. If the search domain would be different depending on which pool your IP address comes from, then it should be added here. if not, then it could live in the templates.

jzhoucliqr: Thanks. Make sense. Will see if there is any use case for per pool domains.

PR #10, opened 2020-07-17
🏃 Add fmuyassarov, jan-est and furkatgofurov7 as reviewers

maelk:

This PR adds some reviewers. Since this repo is fairly new, we cannot expect 10 commits before nominating reviewers. In addition the low number of reviewers might be problematic in this holiday period. Adding some people from the Ericsson team working on Metal3.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -5,3 +5,8 @@ approvers:
  - russellb
  - maelk
  - kashifest
+
+reviewers:
+- fmuyassarov
+- jan-est
+- furkatgofurov7


maelk:
/cc @dhellmann 
/cc @russellb 
is it ok for you ?

dhellmann:
This is fine with me. The 10 reviews/patches line was for setting up the existing repos. When we start new ones, we should include reviewers from the team doing the work.

PR #24, opened 2021-02-22
🏃Rename shortNames with prefix 'm3' or 'metal3'

Insullone:
This is used to avoid naming conflicts with k8s resources' names.

### DIFF
--- a/api/v1alpha1/ipaddress_types.go
+++ b/api/v1alpha1/ipaddress_types.go
@@ -51,7 +51,7 @@ type IPAddressSpec struct {
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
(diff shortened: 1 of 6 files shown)


fmuyassarov (commented):
Looks good. I think you will need to make a couple of changes in other repos that are currently relying on old naming.
Listing some of the places, but might be more:

https://github.com/metal3-io/metal3-dev-env/blob/7ab1eff181df5b9db21ad9eab8e63c04b24c7281/scripts/fetch_manifests.sh#L25

https://github.com/metal3-io/metal3-dev-env/blob/c7c5b33675e780ba899decf566374aede7b1d6b6/scripts/feature_tests/cleanup_env.sh#L44

Insullone:
@fmuyassarov Thanks, I will fix them

maelk (approved):

maelk:
@fmuyassarov There should be backwards compatibility since no shortname was removed. So this should not need any change on other repos

PR #29, opened 2021-04-15
🐛 Fix issue with pre-allocated addresses

fmuyassarov:
In case of pre-allocated addresses, the claims were not given the
proper prefix, gateway and dns server overrides. In addition to
fixing this, some validations on the pool updates are added to
verify that no addresses would be out of bonds.


kashifest (commented):
/lgtm

PR #31, opened 2021-05-02
🏃 Revert Add travis file

fmuyassarov:
This reverts https://github.com/metal3-io/ip-address-manager/pull/27 as Prow cluster is back on track and disables unit & linting test.


kashifest:
This is not needed anymore since we have disabled travis from the settings.

fmuyassarov:
We have disabled Travis instead of deleting the configuration file.
/close

PR #41, opened 2021-07-28
🐛 Fix examples generation

furkatgofurov7:
Fixes examples generation after CAPI uplift to v1a4.
Makes sure to run:

**- make deploy**
**- make deploy-examples**
**- make delete-examples**

Makefile targets successfully.

### DIFF
--- a/examples/ippool/ippool.yaml
+++ b/examples/ippool/ippool.yaml
@@ -1,35 +1,41 @@
 ---
-apiVersion: cluster.x-k8s.io/v1alpha3
+apiVersion: cluster.x-k8s.io/v1alpha4
 kind: Cluster
 metadata:
   name: ${CLUSTER_NAME}
+  namespace: ${NAMESPACE}
 spec:
   clusterNetwork:
     services:
       cidrBlocks: ["10.96.0.0/12"]
     pods:
-      cidrBlocks: ["192.168.0.0/16"]
+      cidrBlocks: ["192.168.0.0/18"]
     serviceDomain: "cluster.local"
   infrastructureRef:
-    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha5
     kind: Metal3Cluster
     name: ${CLUSTER_NAME}
+    namespace: ${NAMESPACE}
   controlPlaneRef:
     kind: KubeadmControlPlane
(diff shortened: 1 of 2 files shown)


examples/ippool/ippool.yaml
```diff
@@ -1,35 +1,40 @@
...
+  namespace: ${NAMESPACE}
 spec:
   clusterName: ${CLUSTER_NAME}
+  namePrefix: ${CLUSTER_NAME}-prov
 ---
 apiVersion: ipam.metal3.io/v1alpha1
 kind: IPClaim
 metadata:
-  name: Claim1
+  name: ${CLUSTER_NAME}-controlplane-template-0-provisioning-pool
```
fmuyassarov, line 36: can you maybe update this example name in https://github.com/metal3-io/ip-address-manager/blob/a1958aab9b77fb571c57e83a9306870f2011ea8a/docs/api.md too?

furkatgofurov7: Sure, done.

furkatgofurov7: Please see #42 I updated the documentation there  since this PR is only for fixing examples generation.

fmuyassarov (commented):
/lgtm

examples/ippool/ippool.yaml
```diff
@@ -1,35 +1,40 @@
...
 metadata:
   name: ${CLUSTER_NAME}
+  namespace: ${NAMESPACE}
 spec:
   clusterNetwork:
     services:
       cidrBlocks: ["10.96.0.0/12"]
     pods:
-      cidrBlocks: ["192.168.0.0/16"]
-    serviceDomain: "cluster.local"
```
kashifest, line 12: I might be missing the context, do we remove this(serviceDomain) also?

kashifest (commented):
/hold
Please unhold if my comment inline is irrelevant.
/approve

examples/ippool/ippool.yaml
furkatgofurov7: Thanks for review @kashifest!  Yes, we do here, since I have not seen `serviceDomain` in the cluster templates nor in the master CI manifests or in PRs'. However, it does not hurt either way if we remove or keep it here, and field is not deprecated or anything, it is still supported in [CAPI](https://github.com/kubernetes-sigs/cluster-api/blob/master/test/infrastructure/docker/examples/simple-cluster.yaml#L19). To be on safe side, reverted it back.

PR #46, opened 2021-08-31
🏃 Add Feruz and Furkat as approvers

kashifest:


russellb:
/approve

We can let @maelk lgtm this

PR #51, opened 2021-09-09
✨ Add AGE column to IPAM CRDs

kashifest:
This adds the AGE column to all our CRDs to make the printed output consistent with the built in K8s resources.

Keeping things consistent with CAPI# https://github.com/kubernetes-sigs/cluster-api/pull/5180


furkatgofurov7 (commented):
/lgtm

PR #52, opened 2021-09-13
🐛 Fix kustomization for IPAM

furkatgofurov7:
This PR fixes an issue in kustomization where cert-manager ca inject annotation for ipam-serving-cert does not contain capm3 namePrefix but the name of the certificate does.


furkatgofurov7:
This issue will be tackled by [capm3#297](https://github.com/metal3-io/cluster-api-provider-metal3/pull/297)

/close

PR #61, opened 2021-10-14
⚠️  Add main branch reference

kashifest:
Since we are changing main as the default branch, we also refer main branch instead of master .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -336,7 +336,7 @@ release-staging: ## Builds and push container images to the staging bucket.
 
 .PHONY: release-tag-latest
 release-tag-latest: ## Adds the latest tag to the last build tag.
-	## TODO(vincepri): Only do this when we're on master.
+	## TODO(vincepri): Only do this when we're on main.
 	gcloud container images add-tag $(CONTROLLER_IMG):$(TAG) $(CONTROLLER_IMG):latest
 
 .PHONY: release-notes
--- a/Makefile
+++ b/Makefile
@@ -331,13 +331,8 @@ release-binary: $(RELEASE_DIR)
 
 .PHONY: release-staging
 release-staging: ## Builds and push container images to the staging bucket.
-	REGISTRY=$(STAGING_REGISTRY) $(MAKE) docker-build-all docker-push-all release-tag-latest
+	REGISTRY=$(STAGING_REGISTRY) $(MAKE) docker-build-all docker-push-all
 
-
-.PHONY: release-tag-latest
-release-tag-latest: ## Adds the latest tag to the last build tag.
-	## TODO(vincepri): Only do this when we're on main.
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -336,7 +336,7 @@ release-staging: ## Builds and push container images to the staging bucket.
 
 .PHONY: release-tag-latest
 release-tag-latest: ## Adds the latest tag to the last build tag.
-	## TODO(vincepri): Only do this when we're on master.
+	## TODO(vincepri): Only do this when we're on main.
 	gcloud container images add-tag $(CONTROLLER_IMG):$(TAG) $(CONTROLLER_IMG):latest
```
fmuyassarov, line 340: These target should probably be removed, we don't use it at all

fmuyassarov (approved):

Makefile
furkatgofurov7: +1

VERSIONING.md
```diff
@@ -33,7 +33,7 @@ but essentially, for any given release X.Y.Z:
 changes.
 
 These guarantees extend to all code exposed in public APIs of
-Cluster API Provider Metal3. This includes code both in Cluster API Provider
+IP Address Manager. This includes code both in Cluster API Provider
```
furkatgofurov7, line 36:
```suggestion
IP Address Manager. This includes code both in IP Address Manager
```

VERSIONING.md
```diff
@@ -80,17 +80,17 @@ separately.
 
 ### Commands and Workflow
 
-Cluster API Provider Metal3 follows the standard Kubernetes workflow: any PR
+IP Address Manager follows the standard Kubernetes workflow: any PR
 needs `lgtm` and `approved` labels, PRs authors must have signed the CNCF CLA,
 and PRs must pass the tests before being merged.  See [the contributor
-docs](https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#the-testing-and-merge-workflow)
+docs](https://github.com/kubernetes/community/blob/main/contributors/guide/pull-requests.md#the-testing-and-merge-workflow)
```
furkatgofurov7, line 86: broken URL, this has not moved to 'main' yet, please revert it

config/default/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: quay.io/metal3-io/ipam:master
+      - image: quay.io/metal3-io/ipam:main
```
furkatgofurov7, line 11: I am a bit confused here, the main tag should already exist in quay.io if I am not wrong and how do we trigger image build with that tag?

furkatgofurov7 (changes requested):

PR #73, opened 2021-11-23
 🐛 Update examples generation after v1beta1 uplift

Sunnatillo:

This PR updates examples generation after v1beta1 uplift

### DIFF
--- a/examples/generate.sh
+++ b/examples/generate.sh
@@ -15,18 +15,15 @@
 
 set -o errexit
 set -o nounset
+set -o pipefail
 
 # Directories.
 SOURCE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 OUTPUT_DIR=${OUTPUT_DIR:-${SOURCE_DIR}/_out}
 
-# Binaries
-ENVSUBST=${ENVSUBST:-envsubst}
-command -v "${ENVSUBST}" >/dev/null 2>&1 || echo -v "Cannot find ${ENVSUBST} in path."
-
 # Cluster.
 export CLUSTER_NAME="${CLUSTER_NAME:-test1}"
-
+export NAMESPACE="${NAMESPACE:-capm3-system}"
 
 # Outputs.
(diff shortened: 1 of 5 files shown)


hack/hack/kind_with_registry.sh
```diff
@@ -0,0 +1,77 @@
...
+  containers=$(docker network inspect ${kind_network} -f "{{range .Containers}}{{.Name}} {{end}}")
+  needs_connect="true"
+  for c in $containers; do
+    if [ "$c" = "${reg_name}" ]; then
+      needs_connect="false"
+    fi
+  done
+  if [ "${needs_connect}" = "true" ]; then
+    docker network connect "${kind_network}" "${reg_name}" || true
+  fi
```
furkatgofurov7, line 77:
```suggestion
if [ "${kind_network}" != "bridge" ]; then
  containers=$(docker network inspect ${kind_network} -f "{{range .Containers}}{{.Name}} {{end}}")
  needs_connect="true"
  for c in $containers; do
    if [ "$c" = "${reg_name}" ]; then
      needs_connect="false"
    fi
  done
  if [ "${needs_connect}" = "true" ]; then
    docker network connect "${kind_network}" "${reg_name}" || true
  fi
fi
```

furkatgofurov7 (changes requested):

examples/generate.sh
```diff
@@ -72,21 +69,27 @@ if [ $OVERWRITE -ne 1 ] && [ -d "$OUTPUT_DIR" ]; then
 fi
 
 mkdir -p "${OUTPUT_DIR}"
+kubectl create namespace "${NAMESPACE}"
+
+# Get enhanced envsubst version to evaluate expressions like ${VAR:=default}
```
furkatgofurov7, line 74: So this is what we were missing in generate.sh and that is why CAPI pod was failing not being able to evaluate some expressions, since we used older version of envsubst.

furkatgofurov7 (approved):

fmuyassarov (commented):
/lgtm

PR #80, opened 2021-12-15
🐛Fix unrelated capm3 mentions under hack folder scripts

furkatgofurov7:
Replaces unrelated capm3 mentions with the proper path in the scripts under the hack folder

### DIFF
--- a/hack/gofmt.sh
+++ b/hack/gofmt.sh
@@ -14,9 +14,9 @@ if [ "${IS_CONTAINER}" != "false" ]; then
 else
   "${CONTAINER_RUNTIME}" run --rm \
     --env IS_CONTAINER=TRUE \
-    --volume "${PWD}:/go/src/github.com/metal3-io/cluster-api-provider-metal3:ro,z" \
+    --volume "${PWD}:/metal3-ipam:ro,z" \
     --entrypoint sh \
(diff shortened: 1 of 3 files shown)


hack/gofmt.sh
```diff
@@ -14,9 +14,9 @@ if [ "${IS_CONTAINER}" != "false" ]; then
 else
   "${CONTAINER_RUNTIME}" run --rm \
     --env IS_CONTAINER=TRUE \
-    --volume "${PWD}:/go/src/github.com/metal3-io/cluster-api-provider-metal3:ro,z" \
+    --volume "${PWD}:/workdir:ro,z" \
     --entrypoint sh \
-    --workdir /go/src/github.com/metal3-io/cluster-api-provider-metal3 \
+    --workdir /workdir \
```
fmuyassarov, line 19: maybe call it ipam? 
```suggestion
    --workdir /ipam \
```

furkatgofurov7: I am not sure that would work, let me try it locally

fmuyassarov: you need to change the mount path as well

fmuyassarov (commented):
/lgtm

kashifest (approved):

PR #100, opened 2022-02-18
✨ Update: Capi to v1.1.2, controller runtime to v0.11.1, cert-manager URL and version to v1.5.3 and Kustomize to v4.4.1

kashifest:


hack/tools/install_kustomize.sh
```diff
@@ -2,7 +2,7 @@
 
 [[ -f bin/kustomize ]] && exit 0
 
-version=3.5.4
+version=4.4.1
```
kashifest, line 5: I will get rid of this file in a followup PR to ensure we dont use two different files for installing kustomize. Since this file is used in CI dockerfiles as well,I am keeping it as it is for now.

lentzi90 (approved):

smoshiur1237 (approved):
LGTM

furkatgofurov7 (commented):
/lgtm

PR #112, opened 2022-03-21
🌱 Ensure setting image tag for current release in infra components

fmuyassarov:
**What this PR does / why we need it:**
We have had a release with infrastructure-components.yaml not updated
to point to the release tag image of IPAM. `MANIFEST_IMG` ensures
to patch the infrastructure-components.yaml before publishing the release
artifacts.


furkatgofurov7 (approved):
/test unit
/approve
/cc @kashifest

PR #119, opened 2022-06-30
 🐛 Don't consider IPAddresses in use when IPPool is being deleted

furkatgofurov7:
Related to: https://github.com/metal3-io/cluster-api-provider-metal3/pull/656

### DIFF
--- a/ipam/ippool_manager.go
+++ b/ipam/ippool_manager.go
@@ -112,8 +112,14 @@ func (m *IPPoolManager) getIndexes(ctx context.Context) (map[ipamv1.IPAddressStr
 
 	addresses := make(map[ipamv1.IPAddressStr]string)
 
-	for _, address := range m.IPPool.Spec.PreAllocations {
-		addresses[address] = ""
+	// After addresses map is populated, we consider that there are still addresses in use.
+	// However, when IPPool.Spec.PreAllocations is given, it can still hold addresses even
+	// though they are already deleted. This consideration should be valid only when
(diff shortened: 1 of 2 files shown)


ipam/ippool_manager.go
```diff
@@ -113,8 +113,14 @@ func (m *IPPoolManager) getIndexes(ctx context.Context) (map[ipamv1.IPAddressStr
...
 
-	for _, address := range m.IPPool.Spec.PreAllocations {
-		addresses[address] = ""
+	// After addresses map is populated, we consider that there are still addresses in use.
+	// However, when IPPool.Spec.PreAllocations is given, it can still hold addresses even
+	// though they are already deleted. This consideration should be valid only when
+	//  PreAllocations left empty.
+	if m.IPPool.DeletionTimestamp.IsZero() && m.IPPool.Spec.PreAllocations == nil {
+		for _, address := range m.IPPool.Spec.PreAllocations {
+			addresses[address] = ""
```
lentzi90, line 122: I don't understand. Here we first check that `PreAllocations == nil`, then we try to do a loop over them? This will never do anything.
Should it be `PreAllocations != nil` instead?

furkatgofurov7: Yes, it was old code here that was left outdated than the local branch I was testing the feature. Should be good now

kashifest (approved):

lentzi90 (approved):

PR #126, opened 2022-08-08
🌱 Improve logging when patching IPClaim

lentzi90:

Log the error when patching fails so we can understand why.

Also moved the patching so we don't try to patch every IPClaim that we delete (it always fails because it is gone). This I'm less sure about but it felt like a good change. Let me know what you think. Perhaps it is better to make sure we *always* patch no matter what happens?


smoshiur1237 (approved):
Thanks for adding the logging improvement.
/lgtm

ipam/ippool_manager.go
```diff
@@ -212,17 +212,17 @@ func (m *IPPoolManager) updateAddress(ctx context.Context,
...
 	addressClaim.Status.ErrorMessage = nil
 
 	if addressClaim.DeletionTimestamp.IsZero() {
+		// Always patch addressClaim exiting this function so we can persist any changes.
+		defer func() {
+			err := helper.Patch(ctx, addressClaim)
+			if err != nil {
+				m.Log.Info("failed to Patch IPClaim", "error", err)
+			}
+		}()
```
fmuyassarov, line 225: this refactoring seems reasonable but I think we should be using `Log.Error` instead of `Log.Info` here

lentzi90: Sure, I can change that! :slightly_smiling_face:

fmuyassarov (approved):

smoshiur1237 (approved):
/lgtm

PR #131, opened 2022-08-30
🌱 Increase linter timeout

lentzi90:

We started seeing timeouts for the linter that we run as part of the unit tests. This PR increases the timeout from 1m to 5m.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,12 +77,12 @@ help:  ## Display this help
 ## Testing
 ## --------------------------------------
 
-.PHONY: testprereqs
-testprereqs: $(KUBEBUILDER) $(KUSTOMIZE)
-
-.PHONY: test
-test: testprereqs generate fmt lint ## Run tests
+.PHONY: unit
+unit: ## Run tests
 	source ./hack/fetch_ext_bins.sh; fetch_tools; setup_envs; go test -v ./controllers/... ./ipam/... -coverprofile ./cover.out; cd $(APIS_DIR); go test -v ./... -coverprofile ./cover.out
+
+.PHONY: test  ## Run formatter, linter and tests
+test: generate fmt lint unit
 
 .PHONY: test-integration
 test-integration: ## Run integration tests
@@ -135,7 +135,7 @@ $(KUSTOMIZE): $(TOOLS_DIR)/go.mod
 
 .PHONY: lint
 lint: $(GOLANGCI_LINT) ## Lint codebase
-	$(GOLANGCI_LINT) run -v
+	$(GOLANGCI_LINT) run -v --timeout=5m
 	cd $(APIS_DIR); ../$(GOLANGCI_LINT) run -v --timeout=10m
 
 lint-full: $(GOLANGCI_LINT) ## Run slower linters to detect possible issues
--- a/hack/unit.sh
+++ b/hack/unit.sh
@@ -10,7 +10,7 @@ if [ "${IS_CONTAINER}" != "false" ]; then
   mkdir /tmp/unit
   cp -r ./* /tmp/unit
   cd /tmp/unit
-  make test
+  make unit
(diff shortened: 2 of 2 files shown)


lentzi90:
Rerunning a few times to make sure we don't hit timeouts
/retest

Makefile
```diff
@@ -80,10 +80,13 @@ help:  ## Display this help
...
 testprereqs: $(KUBEBUILDER) $(KUSTOMIZE)
 
-.PHONY: test
-test: testprereqs generate fmt lint ## Run tests
+.PHONY: unit
+unit: testprereqs ## Run tests
 	source ./hack/fetch_ext_bins.sh; fetch_tools; setup_envs; go test -v ./controllers/... ./ipam/... -coverprofile ./cover.out; cd $(APIS_DIR); go test -v ./... -coverprofile ./cover.out
 
+.PHONY: test  ## Run formatter, linter and tests
+test: fmt lint unit
```
kashifest, line 88: I would still want to keep test as is i.e.  Keep generate

mboukhalfa:
> **What this PR does / why we need it**:
> 
> We started seeing timeouts for the linter that we run as part of the unit tests. This PR increases the timeout from 1m to 2m.

the code change shows --timeout=5m

Makefile
lentzi90: Good point. Fixed!

Makefile
```diff
@@ -80,10 +80,13 @@ help:  ## Display this help
...
 testprereqs: $(KUBEBUILDER) $(KUSTOMIZE)
 
-.PHONY: test
-test: testprereqs generate fmt lint ## Run tests
+.PHONY: unit
+unit: testprereqs ## Run tests
 	source ./hack/fetch_ext_bins.sh; fetch_tools; setup_envs; go test -v ./controllers/... ./ipam/... -coverprofile ./cover.out; cd $(APIS_DIR); go test -v ./... -coverprofile ./cover.out
 
+.PHONY: test  ## Run formatter, linter and tests
+test: generate fmt lint unit
```
furkatgofurov7, line 88: I am wondering if we need `generate` for test target, at least in CAPM3 we are not doing it https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/Makefile#L116-L117. It makes sense to have it though so maybe we should update the CAPM3 makefile "test" target

Makefile
```diff
@@ -80,10 +80,13 @@ help:  ## Display this help
 .PHONY: testprereqs
 testprereqs: $(KUBEBUILDER) $(KUSTOMIZE)
 
-.PHONY: test
-test: testprereqs generate fmt lint ## Run tests
+.PHONY: unit
+unit: testprereqs ## Run tests
```
furkatgofurov7, line 84: do you need `testprereqs` ?

lentzi90: Hmm I thought we needed it for kubebuilder but looks like that is already taken care of with the scripts... But if we don't need testprereqs for the tests, what do we use it for? Is it for generate?

Makefile
lentzi90: I agree we should add this in CAPM3 if it isn't too heavy to run all the time.

Makefile
kashifest: Wouldnt it be needed for local tests? I am not sure if removing it is a good idea .

furkatgofurov7: No, that was removed from CAPM3 too some time ago, xref: https://github.com/metal3-io/cluster-api-provider-metal3/pull/389

kashifest: ok

furkatgofurov7 (commented):
I am seeing much of this happening in https://github.com/metal3-io/ip-address-manager/pull/119 and this should help with timeout issue in unit tests. 

/lgtm

PR #132, opened 2022-08-30
🌱 Improve logging when releasing address

lentzi90:

- Add log when clearing IPClaim from IPPools allocations.
- Disable logs for unit tests

### DIFF
--- a/controllers/ippool_controller_test.go
+++ b/controllers/ippool_controller_test.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 
+	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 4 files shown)


furkatgofurov7 (approved):

kashifest:
> * Disable logs for unit tests

Does this mean there wont be any log while you run unit tests, wouldn't it be difficult to debug while writing tests ?

lentzi90:
> > * Disable logs for unit tests
> 
> Does this mean there wont be any log while you run unit tests, wouldn't it be difficult to debug while writing tests ?

It is the same we do in CAPM3 and other places (e.g. [here](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/baremetal/metal3cluster_manager_test.go#L82). Logs are generally not useful in unit tests since we run hundreds of them and often in parallel. This makes it hard to keep track of which test the logs are coming from. Nothing is preventing enabling the logs again for specific tests while debugging.

kashifest:
> parallel

Thanks for the clarification. I am still old school and like lots of logs!!!!!

/approve

PR #133, opened 2022-09-01
🌱 Use finalizers instead of ownerRefs to check if IPClaim is in use

lentzi90:

The previous implementation was API breaking, since it required the
user/controller owning the IPClaim to remove the ownerRef before the
IPAddress would be deleted. This commit changes it so that we check for
finalizers instead. More specifically, if there are any other finalizers
than the IPAM one, we consider the IPClaim to be still in use. In this
way we "revert" the API breaking change but still get the same
functionality.

### DIFF
--- a/ipam/ippool_manager.go
+++ b/ipam/ippool_manager.go
@@ -224,11 +224,12 @@ func (m *IPPoolManager) updateAddress(ctx context.Context,
 			return addresses, err
 		}
 	} else {
-		// Check if this claim is in use. Does it have any owners?
+		// Check if this claim is in use. Does it have any other finalizers than our own?
 		// If it is no longer in use, proceed to delete the associated IPAddress
-		if len(addressClaim.OwnerReferences) > 0 {
-			m.Log.Info("IPClaim is still in use (has owners). Cannot delete IPAddress.",
-				"IPClaim", addressClaim.Name, "Owners", addressClaim.OwnerReferences)
+		if len(addressClaim.Finalizers) > 1 ||
+			(len(addressClaim.Finalizers) == 1 && !Contains(addressClaim.Finalizers, ipamv1.IPClaimFinalizer)) {
+			m.Log.Info("IPClaim is still in use (has other finalizers). Cannot delete IPAddress.",
+				"IPClaim", addressClaim.Name, "Finalizers", addressClaim.Finalizers)
 			return addresses, nil
 		}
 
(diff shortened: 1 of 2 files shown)


kashifest (approved):

ipam/ippool_manager.go
```diff
@@ -224,10 +224,11 @@ func (m *IPPoolManager) updateAddress(ctx context.Context,
...
 	} else {
-		// Check if this claim is in use. Does it have any owners?
+		// Check if this claim is in use. Does it have any other finalizers than our own?
 		// If it is no longer in use, proceed to delete the associated IPAddress
-		if len(addressClaim.OwnerReferences) > 0 {
-			m.Log.Info("IPClaim is still in use (has owners). Cannot delete IPAddress.",
+		if len(addressClaim.Finalizers) > 1 ||
+			(len(addressClaim.Finalizers) == 1 && !Contains(addressClaim.Finalizers, ipamv1.IPClaimFinalizer)) {
+			m.Log.Info("IPClaim is still in use (has other finalizers). Cannot delete IPAddress.",
 				"IPClaim", addressClaim.Name, "Owners", addressClaim.OwnerReferences)
```
furkatgofurov7, line 232:
```suggestion
				"IPClaim", addressClaim.Name, "Finalizers", addressClaim.Finalizers)
```

ipam/ippool_manager.go
```diff
@@ -224,10 +224,11 @@ func (m *IPPoolManager) updateAddress(ctx context.Context,
 			return addresses, err
 		}
 	} else {
-		// Check if this claim is in use. Does it have any owners?
+		// Check if this claim is in use. Does it have any other finalizers than our own?
 		// If it is no longer in use, proceed to delete the associated IPAddress
-		if len(addressClaim.OwnerReferences) > 0 {
-			m.Log.Info("IPClaim is still in use (has owners). Cannot delete IPAddress.",
+		if len(addressClaim.Finalizers) > 1 ||
+			(len(addressClaim.Finalizers) == 1 && !Contains(addressClaim.Finalizers, ipamv1.IPClaimFinalizer)) {
```
furkatgofurov7, line 230: you could do:
```suggestion
		if len(addressClaim.Finalizers) >= 1 && !Contains(addressClaim.Finalizers, ipamv1.IPClaimFinalizer)) {
```

furkatgofurov7 (changes requested):

furkatgofurov7 (commented):
/lgtm

PR #153, opened 2022-10-03
🌱Add CAPI metadata.yaml and add namespace resource in release manifests

jessehu:
Add CAPI metadata.yaml and add namespace resource in release manifests
- metadata.yaml is needed by CAPI clusterctl.
- namespace capm3-system is needed for deploying ip-address-manager.

Refer to CAPI files:
- https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.2.2
- https://github.com/kubernetes-sigs/cluster-api/blob/v1.2.2/Makefile#L748
- https://github.com/kubernetes-sigs/cluster-api/blob/v1.2.2/config/default/kustomization.yaml#L8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -309,6 +309,8 @@ $(RELEASE_NOTES_DIR):
 .PHONY: release-manifests
 release-manifests: $(KUSTOMIZE) $(RELEASE_DIR) ## Builds the manifests to publish with a release
 	$(KUSTOMIZE) build config/default > $(RELEASE_DIR)/ipam-components.yaml
+	# Add metadata to the release artifacts
+	cp metadata.yaml $(RELEASE_DIR)/metadata.yaml
 
 .PHONY: release-notes
 release-notes: $(RELEASE_NOTES_DIR) $(RELEASE_NOTES)
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -10,6 +10,9 @@ namePrefix: ipam-
 commonLabels:
   cluster.x-k8s.io/provider: "infrastructure-metal3"
 
+resources:
+- namespace.yaml
+
 bases:
 - ../rbac
 - ../manager
--- a/config/default/namespace.yaml
+++ b/config/default/namespace.yaml
@@ -1,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: system
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -1,0 +1,20 @@
+# maps release series of major.minor to cluster-api contract version
+# the contract version may change between minor or major versions, but *not*
+# between patch versions.
+#
+# update this file only when a new major or minor version is released
+apiVersion: clusterctl.cluster.x-k8s.io/v1alpha3
+kind: Metadata
+releaseSeries:
+  - major: 1
+    minor: 4
+    contract: v1alpha1
+  - major: 1
+    minor: 3
+    contract: v1alpha1
+  - major: 1
+    minor: 2
+    contract: v1alpha1
+  - major: 1
+    minor: 1
+    contract: v1alpha1


jessehu:
The change of ipam-components.yaml is:
```
diff out/ipam-components.yaml path/to/ip-address-manager/v1.2.0/ipam-components.yaml
1,8d0
< apiVersion: v1
< kind: Namespace
< metadata:
<   labels:
<     cluster.x-k8s.io/provider: infrastructure-metal3
<     control-plane: controller-manager
<   name: capm3-system
< ---
```

kashifest:
> * namespace capm3-system is needed for deploying ip-address-manager.

So how are you planning to deploy it, not as part of the provider ?

jessehu:
hi @kashifest we deploy it as a CAPI provider.

```
$ ls -1 providers/infrastructure-ipam/v1.2.1
ipam-components.yaml
metadata.yaml

$ head providers/infrastructure-ipam/v1.2.1/ipam-components.yaml
apiVersion: v1
kind: Namespace
metadata:
  labels:
    control-plane: controller-manager
  name: capm3-system
---
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
```

jessehu:
ping there!
is this ok to merge ?

jessehu:
Thanks @mboukhalfa. I rebased the main branch today.

tuminoid:
> Thanks @mboukhalfa. I rebased the main branch today.

Please also squash the commits.

/retest

jessehu:
commits squashed

tuminoid:
This is failing tests. Added namepsace is colliding with existing namespace.

```console
hack/tools/bin/kustomize build config/default > out/infrastructure-components.yaml
Error: accumulating resources: accumulation err='accumulating resources from '../ipam': '/home/****/metal3/config/ipam' must resolve to a file': recursed merging from path '/home/****/metal3/config/ipam': may not add resource with an already registered id: ~G_v1_Namespace|~X|capm3-system
make[2]: *** [Makefile:507: release-manifests] Error 1
```

lentzi90 (commented):
IPAM is currently deployed as part of CAPM3, which means that the namespace and metadata will all come from CAPM3. We can certainly discuss if that is how it *should* be, but the point is that these changes would break that without more changes in other places. This is why the integration tests fail.

If you check the [CAPM3 infrastructure-components.yaml](https://github.com/metal3-io/cluster-api-provider-metal3/releases/download/v1.4.0/infrastructure-components.yaml), you will find that it includes IPAM (and also the namespace). So adding the namespace here is causing a conflict.

I see a few options here for how to move forward. One is to create a "standalone" kustomization where we add the namespace so the "default" stays as it is. Another option would be to make some changes in CAPM3 so that it for example doesn't use the "default" kustomization. We could also discuss if CAPM3 should include IPAM at all. This would need proper backing from the community though as it is definitely a breaking change.

jessehu:
This is the output in my local box:
```
ip-address-manager (huhui/add-metadata) $ make release-manifests
cd hack/tools; ./install_kustomize.sh
+ [[ -f bin/kustomize ]]
+ exit 0
hack/tools/bin/kustomize build config/default > out/ipam-components.yaml
# Add metadata to the release artifacts
cp metadata.yaml out/metadata.yaml
```
The e2e error log says "> out/infrastructure-components.yaml". It's a different yaml file.

lentzi90:
@jessehu please see my comment above. We do not deploy IPAM separate from CAPM3 in the CI or in our own environments. It is expected to be deployed via `clusterctl init --infrastructure metal3`. I'm glad to discuss changes to this, but as things are, your changes here is breaking the "normal" deploy procedure. That is why the tests fail.

jessehu:
Thanks @lentzi90. Didn't see your comment before I sent out mine. Another option would be removing the namepsace piece in my PR?

lentzi90:
Sure! I thought you needed it though?
Can you explain how you deploy IPAM btw+ I saw the [comment](https://github.com/metal3-io/ip-address-manager/pull/153#issuecomment-1341782680) before, but I am not understanding from it how you do it. Sounds like you use clusterctl in some way. Do you use IPAM with another infrastructure provider or completely separate from CAPI? :thinking: 
It would probably be easier to understand what makes sense if you explain why this is needed :slightly_smiling_face:

lentzi90:
Great! This is something I think we should support in a better way :+1:

lentzi90:
@kashifest this is related to making IPAM a real IPAM provider for CAPI, as we talked about, but as you can see we would need to change CAPM3 to make the e2e tests pass. Now we get conflicts because CAPM3 already includes IPAM :slightly_frowning_face:

tuminoid:
/retest
New "make test" is working now.

kashifest:
> @kashifest this is related to making IPAM a real IPAM provider for CAPI, as we talked about, but as you can see we would need to change CAPM3 to make the e2e tests pass. Now we get conflicts because CAPM3 already includes IPAM 🙁

Yes, we need to separate out IPAM from CAPM3, it is now deployed as a sub component of infra, it needs to be a ipam component rather. I will look into it.

lentzi90:
/remove-lifecycle stale
@kashifest is working on this

PR #195, opened 2023-01-17
✨ add explicit securitycontexts to controller

tuminoid:
It is always good to not rely on the defaults, but be explicit. Set explicit, secure securityContext for the IPAM controller manager deployment and containers.

CAPI has the same starting from upcoming v1.4.0 and cert-manager etc has them already.

Also, reindent manifest properly.


lentzi90 (commented):
/lgtm

kashifest (approved):

PR #199, opened 2023-02-14
🐛 hack: gomod needs to verify all go.mod files

tuminoid:
gomod.sh has only checked top-level go.mod/go.sum so far, which means api/ and hack/tools go.mods have not been verified.

The script also did not work locally at all, since /workdir was mounted read-only and "go mod tidy" writes to the directory.

Git also has new safe.directory setting which gets applies when running locally, and breaks all Git operations unless we configure it to ignore the work dir.


kashifest (approved):

kashifest:
Giving some time for others to review.

tuminoid:
Fixed wrong indentation.

/test-ubuntu-integration-main

tuminoid:
/cc @lentzi90 @furkatgofurov7 

Your opinions on this rewrite, please. Same will be applied to BMO and CAPM3 after we agree on this one on the approach.

hack/gomod.sh
```diff
@@ -5,44 +5,63 @@
...
+        echo "Dirty tree: refusing to continue out of caution"
+        exit 1
+    fi
 
-  if [ $rc -ne 0 ]; then
-      echo "'go mod tidy' failed"
-      exit 1;
-  fi
+    # sh doesn't allow ** recursive glob
+    for mod_file in *.mod */*.mod */*/*.mod; do
```
furkatgofurov7, line 34: Have you thought of a taking advantage of using https://github.com/metal3-io/ip-address-manager/blob/main/Makefile#L160-L164 target ?

hack/gomod.sh
```diff
@@ -5,44 +5,63 @@
...
-  fi
+        if ! go mod tidy; then
+            echo "'go mod tidy' failed"
+            exit 1
+        fi
 
-  exit 0;
+        STATUS="$(git status --porcelain go.mod go.sum)"
+        if [ -n "${STATUS}" ]; then
+            echo "go.mod and/or go.sum changed"
```
lentzi90, line 50: I see that we echo the dir above, but maybe we could add it here to make it even more obvious?
```suggestion
            echo "go.mod and/or go.sum changed in ${mod_dir}"
```

hack/gomod.sh
lentzi90: I was thinking the same also. In CAPM3 we use make targets in these scripts (e.g. [here](https://github.com/metal3-io/cluster-api-provider-metal3/blob/b640d55d7a3027c6b76bc3affd2c05b6b51eacce/hack/unit.sh#L13)). The glob here would obviously handle it nicely if we add more modules, but we could easily forget to update the Makefile.

tuminoid: Good point, I missed the fact the make actually works in the golang image. Converted to use the make target now.

hack/gomod.sh
tuminoid: I made it output the `git status` output so it prints out all of the changed files.

kashifest:
This look much more clear and effective now. Thanks @tuminoid and the reviewers for pointing this out.

tuminoid:
Good suggestions, implemented. Simplified the code as well.

/test-ubuntu-integration-main

furkatgofurov7 (commented):
Thanks @tuminoid!

/lgtm

tuminoid:
Updated the commit message to reflect new design. @furkatgofurov7 re-add lgtm.

lentzi90 (commented):
/lgtm

furkatgofurov7 (commented):
Small typo in the commit message

/lgtm

PR #226, opened 2023-04-03
🌱 cleanup hack/ scripts

tuminoid:
Cleanup syntax, formatting, indentation etc for scripts in hack. They should be exact copies in each repo (besides if repo specific configuration/files), not a little bit different everywhere.


lentzi90 (commented):
/approve

tuminoid:
Isn't it nice when Tide is actually working again :)

PR #238, opened 2023-05-03
✨ Add flags for configuring rate limits

lentzi90:

Matching https://github.com/kubernetes-sigs/cluster-api/pull/8579

There is a built-in rate limit for the controller when talking to the Kubernetes API. So far it has not been possible to configure this limit. This commit adds flags for setting both the QPS and the burst for the rate limit. The default remains the same as before (20 QPS, 30 burst).

New flags:

--kube-api-qps
--kube-api-burst

Also adds .devcontainer to .gitignore

Fixes #

### DIFF
--- a/main.go
+++ b/main.go
@@ -46,6 +46,8 @@ var (
 	enableLeaderElection bool
 	syncPeriod           time.Duration
 	ippoolConcurrency    int
+	restConfigQPS        float64
+	restConfigBurst      int
 	webhookPort          int
 	healthAddr           string
 	watchNamespace       string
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -99,7 +107,10 @@ func main() {
 	// klog.Background will automatically use the right logger.
 	ctrl.SetLogger(klog.Background())
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	restConfig := ctrl.GetConfigOrDie()
+	restConfig.QPS = float32(restConfigQPS)
```
tuminoid, line 111: RestConfigQPS is defined as float64 on L49, why do we convert it to float32?

lentzi90: Good question. For some reason the QPS field in the rest config is float32 so that is what we need. However, the `flag` library doesn't have a `Float32Var`, only `Float64Var`, so I went with a float64 variable that I convert.

In CAPI and CAPM3 we use another flag library on top of `flag` that *has* a `Float32Var` though. So there we [didn't have to do the conversion](https://github.com/Nordix/cluster-api/blob/ba34136c11580c8cba752a3dd1a408cd8b2a7769/bootstrap/kubeadm/main.go#L128).

lentzi90:
CC @furkatgofurov7 @kashifest

kashifest (approved):

PR #248, opened 2023-05-11
:sparkles: Add TLS configuration flags for IPAM

as20203:
This PR adds TLS configuration flags for IPAM.

This enables flags --tls-min-version and --tls-max-version flags, which can be used to limit TLS versions server accepts. Choices are TLS12 and TLS13, so you can force TLS 1.2 only, TLS 1.3 only, or (default) TLS 1.2 and TLS 1.3 is negotiated between server and client.

This also disables TLS 1.0 and TLS 1.1 as they're insecure and obsolete and not supported by Go TLS library.

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,9 +18,11 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	ipamv1 "github.com/metal3-io/ip-address-manager/api/v1alpha1"
@@ -30,6 +32,7 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	cliflag "k8s.io/component-base/cli/flag"
 	"k8s.io/component-base/logs"
 	logsv1 "k8s.io/component-base/logs/api/v1"
 	"k8s.io/klog/v2"
@@ -38,6 +41,20 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	// +kubebuilder:scaffold:imports
 )
+
+type TLSVersion string
+
+// Constants for TLS versions.
+const (
+	TLSVersion12 TLSVersion = "TLS12"
+	TLSVersion13 TLSVersion = "TLS13"
+)
+
+type TLSOptions struct {
+	TLSMaxVersion   string
+	TLSMinVersion   string
+	TLSCipherSuites string
+}
 
 var (
 	myscheme             = runtime.NewScheme()
@@ -54,6 +71,8 @@ var (
 	webhookCertDir       string
 	watchFilterValue     string
 	logOptions           = logs.NewOptions()
+	tlsOptions           = TLSOptions{}
+	tlsSupportedVersions = []string{"TLS12", "TLS13"}
 )
 
 func init() {
@@ -96,7 +115,23 @@ func main() {
 
 	flag.IntVar(&restConfigBurst, "kube-api-burst", 30,
 		"Maximum number of queries that should be allowed in one burst from the controller client to the Kubernetes API server. Default 30")
+	flag.StringVar(&tlsOptions.TLSMinVersion, "tls-min-version", "TLS12",
+		"The minimum TLS version in use by the webhook server.\n"+
+			fmt.Sprintf("Possible values are %s.", strings.Join(tlsSupportedVersions, ", ")),
+	)
 
+	flag.StringVar(&tlsOptions.TLSMaxVersion, "tls-max-version", "TLS13",
+		"The maximum TLS version in use by the webhook server.\n"+
+			fmt.Sprintf("Possible values are %s.", strings.Join(tlsSupportedVersions, ", ")),
+	)
+
+	tlsCipherPreferredValues := cliflag.PreferredTLSCipherNames()
+	tlsCipherInsecureValues := cliflag.InsecureTLSCipherNames()
+	flag.StringVar(&tlsOptions.TLSCipherSuites, "tls-cipher-suites", "",
+		"Comma-separated list of cipher suites for the webhook server. "+
+			"If omitted, the default Go cipher suites will be used. \n"+
+			"Preferred values: "+strings.Join(tlsCipherPreferredValues, ", ")+". \n"+
+			"Insecure values: "+strings.Join(tlsCipherInsecureValues, ", ")+".")
 	flag.Parse()
 
 	if err := logsv1.ValidateAndApply(logOptions, nil); err != nil {
@@ -106,6 +141,11 @@ func main() {
 
 	// klog.Background will automatically use the right logger.
 	ctrl.SetLogger(klog.Background())
+	tlsOptionOverrides, err := GetTLSOptionOverrideFuncs(tlsOptions)
+	if err != nil {
+		setupLog.Error(err, "unable to add TLS settings to the webhook server")
+		os.Exit(1)
+	}
 
 	restConfig := ctrl.GetConfigOrDie()
 	restConfig.QPS = float32(restConfigQPS)
@@ -121,6 +161,7 @@ func main() {
 		HealthProbeBindAddress:     healthAddr,
 		Namespace:                  watchNamespace,
 		CertDir:                    webhookCertDir,
+		TLSOpts:                    tlsOptionOverrides,
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
@@ -184,4 +225,78 @@ func setupWebhooks(mgr ctrl.Manager) {
 
 func concurrency(c int) controller.Options {
 	return controller.Options{MaxConcurrentReconciles: c}
+}
+
+// GetTLSOptionOverrideFuncs returns a list of TLS configuration overrides to be used
+// by the webhook server.
+func GetTLSOptionOverrideFuncs(options TLSOptions) ([]func(*tls.Config), error) {
+	var tlsOptions []func(config *tls.Config)
+
+	tlsMinVersion, err := GetTLSVersion(options.TLSMinVersion)
+	if err != nil {
+		return nil, err
+	}
+
+	tlsMaxVersion, err := GetTLSVersion(options.TLSMaxVersion)
+	if err != nil {
+		return nil, err
+	}
+
+	if tlsMaxVersion != 0 && tlsMinVersion > tlsMaxVersion {
+		return nil, fmt.Errorf("TLS version flag min version (%s) is greater than max version (%s)",
+			options.TLSMinVersion, options.TLSMaxVersion)
+	}
+
+	tlsOptions = append(tlsOptions, func(cfg *tls.Config) {
+		cfg.MinVersion = tlsMinVersion
+	})
+
+	tlsOptions = append(tlsOptions, func(cfg *tls.Config) {
+		cfg.MaxVersion = tlsMaxVersion
+	})
+	// Cipher suites should not be set if empty.
+	if options.TLSMinVersion == string(TLSVersion13) &&
+		options.TLSMaxVersion == string(TLSVersion13) &&
+		options.TLSCipherSuites != "" {
+		setupLog.Info("warning: Cipher suites should not be set for TLS version 1.3. Ignoring ciphers")
+		options.TLSCipherSuites = ""
+	}
+
+	if options.TLSCipherSuites != "" {
+		tlsCipherSuites := strings.Split(options.TLSCipherSuites, ",")
+		suites, err := cliflag.TLSCipherSuites(tlsCipherSuites)
+		if err != nil {
+			return nil, err
+		}
+
+		insecureCipherValues := cliflag.InsecureTLSCipherNames()
+		for _, cipher := range tlsCipherSuites {
+			for _, insecureCipherName := range insecureCipherValues {
+				if insecureCipherName == cipher {
+					setupLog.Info(fmt.Sprintf("warning: use of insecure cipher '%s' detected.", cipher))
+				}
+			}
+		}
+		tlsOptions = append(tlsOptions, func(cfg *tls.Config) {
+			cfg.CipherSuites = suites
+		})
+	}
+
+	return tlsOptions, nil
+}
+
+// GetTLSVersion returns the corresponding tls.Version or error.
+func GetTLSVersion(version string) (uint16, error) {
+	var v uint16
+
+	switch version {
+	case string(TLSVersion12):
+		v = tls.VersionTLS12
+	case string(TLSVersion13):
+		v = tls.VersionTLS13
+	default:
+		return 0, fmt.Errorf("unexpected TLS version %q (must be one of: TLS12, TLS13)", version)
+	}
+
+	return v, nil
 }
--- a/main_test.go
+++ b/main_test.go
@@ -1,0 +1,111 @@
+/*
+Copyright 2023 The Metal3 Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"bytes"
+	"testing"
+
+	. "github.com/onsi/gomega"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+func TestTLSInsecureCiperSuite(t *testing.T) {
+	t.Run("test insecure cipher suite passed as TLS flag", func(t *testing.T) {
+		g := NewWithT(t)
+		tlsMockOptions := TLSOptions{
+			TLSMaxVersion:   "TLS13",
+			TLSMinVersion:   "TLS12",
+			TLSCipherSuites: "TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256",
+		}
+		ctrl.Log.WithName("setup")
+		ctrl.SetLogger(klog.Background())
+
+		bufWriter := bytes.NewBuffer(nil)
+		klog.SetOutput(bufWriter)
+		klog.LogToStderr(false) // this is important, because klog by default logs to stderr only
+		_, err := GetTLSOptionOverrideFuncs(tlsMockOptions)
+		g.Expect(err).Should(BeNil())
+		g.Expect(bufWriter.String()).Should(ContainSubstring("use of insecure cipher 'TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256' detected."))
+	})
+}
+
+func TestTLSMinAndMaxVersion(t *testing.T) {
+	t.Run("should fail if TLS min version is greater than max version.", func(t *testing.T) {
+		g := NewWithT(t)
+		tlsMockOptions := TLSOptions{
+			TLSMaxVersion: "TLS12",
+			TLSMinVersion: "TLS13",
+		}
+		_, err := GetTLSOptionOverrideFuncs(tlsMockOptions)
+		g.Expect(err.Error()).To(Equal("TLS version flag min version (TLS13) is greater than max version (TLS12)"))
+	})
+}
+
+func Test13CipherSuite(t *testing.T) {
+	t.Run("should reset ciphersuite flag if TLS min and max version are set to 1.3", func(t *testing.T) {
+		g := NewWithT(t)
+
+		// Here TLS_RSA_WITH_AES_128_CBC_SHA is a tls12 cipher suite.
+		tlsMockOptions := TLSOptions{
+			TLSMaxVersion:   "TLS13",
+			TLSMinVersion:   "TLS13",
+			TLSCipherSuites: "TLS_RSA_WITH_AES_128_CBC_SHA,TLS_AES_256_GCM_SHA384",
+		}
+
+		ctrl.Log.WithName("setup")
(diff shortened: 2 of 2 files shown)


as20203:
@kashifest @lentzi90 @tuminoid @mboukhalfa @adilGhaffarDev I've added TLS configuration flags for max-version, min-version and cipher-suites in this PR. There are two more flags regarding private-key-file and tls-cert-file. However, I was unable to find anything about those options in the TLS config struct and the examples in the description PR. Do you guys have any idea how can we add those.

util/flags/tls.go
lentzi90, line 1: I think it would be better to just keep this in `main.go`. At least `AddTLSOptions` can be done inline with the rest of the flags to keep things simple, so no need for a separate function for this. `GetTLSOptionOverrideFuncs` is fine to have as a separate function but I would avoid having a separate package just for this.

lentzi90 (commented):
/approve

tuminoid (changes requested):
I'd like to see some tests added as well.

as20203:
> I'd like to see some tests added as well.

For the tests what type of tests are expected to added? Meaning should we show that the value of flags are being set for the variable after we pass in the value? I couldn't find any existing tests for the other flags so wanted to clarify first.

tuminoid:
> > I'd like to see some tests added as well.
> 
> For the tests what type of tests are expected to added? Meaning should we show that the value of flags are being set for the variable after we pass in the value? I couldn't find any existing tests for the other flags so wanted to clarify first.

For example, we should be failing if we pass TLS 1.2 cipher suites etc flags but specify only TLS 1.3 since 1.3 doesn't have configurable suites. We could also go with warning user when using incompatible flags, but current way of accepting but not using supplied flags is not ok IMO.

We don't need to test parsing the flags otherwise, hence we're not having tests for that, but these conflicting flags need to be addressed somehow.

as20203:
@tuminoid @lentzi90 I've got a few question regarding test cases. 
- As for IPAM I saw that mostly the test files are located in specific folders for example ipam. The test files are inside a specific package like for ipam test cases they belong to ipam package. 
  ### My scenario:
  Now since I'm testing the flags for TLS and we have all the functions directly inside main.go so how will I be able to access 
  those functions in a test file since the functions that are called inside the test file need to be in the same location and package as per my limited knowledge of go. Adding **package main** for a test file located at a different location than **main.go** seemed to give me some error. 
I made a **main_test.go** file in the same location as **main.go** and both belonged to package **main** and then I was able to access them but this will mess up the folder structure of the project I believe. I think that those functions should be in a separate package of its own as as done earlier in this PR and called in the main.go. This way they can be tested as its done with other files. I would like to hear your opinion on this.

- How do I test and run those test cases. I tried go test command as was done in the MakeFile but I get below errors

  ```
  go test main_test.go
  # command-line-arguments [command-line-arguments.test]
  ./main_test.go:35:21: undefined: TLSOptions
  ./main_test.go:40:13: undefined: GetTLSOptionOverrideFuncs
  FAIL    command-line-arguments [build failed]
  FAIL
  ```

  ```
  go test main_test.go main.go
  ok      command-line-arguments  0.043s
  ```
  In the first scenario the test does not run and in the second one I believe that it doesn't seem to detect the test file. In the 
  Makefile its run like below but for my test it doesn't seem to run correctly.
  
  `go test -v ./controllers/...`

lentzi90:
> I made a main_test.go file in the same location as main.go and both belonged to package main and then I was able to access them but this will mess up the folder structure of the project I believe.

I would say this is the way to go. It is the standard way of adding tests in go, i.e. a file with `*_test.go` name right next to the code that is tested and in the same package. The compiler also knows about this and will exclude the test code when making the binaries.

as20203:
@tuminoid I've added two tests, 
- One of them checks for insecure cipher suite passed in flags and fails in case we provide those. 
- The other checks if  tls min and max version are having correct values to avoid cases where min version value is greater than the max version. 
### My Question related to below reply


### Queries related to your reply

> > > I'd like to see some tests added as well.
> > 
> > 
> > For the tests what type of tests are expected to added? Meaning should we show that the value of flags are being set for the variable after we pass in the value? I couldn't find any existing tests for the other flags so wanted to clarify first.
> 
> For example, we should be failing if we pass TLS 1.2 cipher suites etc flags but specify only TLS 1.3 since 1.3 doesn't have configurable suites. We could also go with warning user when using incompatible flags, but current way of accepting but not using supplied flags is not ok IMO.
> 
> We don't need to test parsing the flags otherwise, hence we're not having tests for that, but these conflicting flags need to be addressed somehow.

- There isn't any specific function which can help us get the specific cipher suites for any particular tls version. Either we get all cipher suites using  **cliflag.TLSCipherPossibleValues()** which can be passed to the provided flag as comma separated values so there isn't any function I could find that would get the 1.2 and 1.3 ciphers separately. This article provides some details about which are tls 1.2 ciphers and which are 1.3 ciphers.  [https://crashtest-security.com/configure-ssl-cipher-order](article)|
- Now my question is how should the test case be made since we specify a min-version and a max-version for tls rather than a single specific version we would be using, so should our code allow cipher suites for both the versions and and all those in between that we specify like if min-version is 1.2 and max is 1.3 so should we allow all 1.2 and 1.3 cipher suites? Other than that, based upon above article should I create arrays with cipher suites for both 1.2 and 1.3 versions and then check in the code in case we allow for ciphersuites for all versions include the min and max versions?

main_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+func TestTLSMinAndMaxVersion(t *testing.T) {
+	t.Run("should fail if tls min version is greater than max version.", func(t *testing.T) {
+		g := NewWithT(t)
+		tlsMockOptions := TLSOptions{
+			TLSMaxVersion: "VersionTLS12",
+			TLSMinVersion: "VersionTLS13",
+		}
+		_, err := GetTLSOptionOverrideFuncs(tlsMockOptions)
+		g.Expect(err.Error()).To(Equal("cipher suite flag min version is greater or equal to max version"))
```
tuminoid, line 46: Min and Max can be the same, that shouldn't be an error. Min cannot be higher than Max though.

main_test.go
tuminoid, line 1: What I also would like to see is a test that you can't specify TLS12 cipher suites etc with min version being TLS13.

tuminoid (commented):
/cc @lentzi90 @mboukhalfa 

Would you help with the Go code review?

main_test.go
as20203: added a test case for this scenario.

main_test.go
as20203: fixed this issue.

main.go
```diff
@@ -185,3 +223,85 @@ func setupWebhooks(mgr ctrl.Manager) {
...
+			return nil, err
+		}
+
+		if options.TLSMinVersion == "VersionTLS13" {
+			tls13Suites := []string{"TLS_AES_128_GCM_SHA256", "TLS_AES_256_GCM_SHA384", "TLS_CHACHA20_POLY1305_SHA256"}
+
+			for _, cipherSuite := range tlsCipherSuites {
+				_, containstls13Suite := contains(tls13Suites, cipherSuite)
+				if !containstls13Suite {
+					return nil, fmt.Errorf("invalid cipher suite provided for tls version 13")
```
tuminoid, line 280:
```suggestion
					return nil, fmt.Errorf("invalid cipher suite provided for TLS version 1.3")
```

main.go
```diff
@@ -185,3 +223,85 @@ func setupWebhooks(mgr ctrl.Manager) {
...
+// by the webhook server.
+func GetTLSOptionOverrideFuncs(options TLSOptions) ([]func(*tls.Config), error) {
+	var tlsOptions []func(config *tls.Config)
+	TLSMinVersionIndex, _ := contains(cliflag.TLSPossibleVersions(), options.TLSMinVersion)
+	TLSMaxVersionIndex, _ := contains(cliflag.TLSPossibleVersions(), options.TLSMaxVersion)
+	// Check if tls min version flag value is greater than max version
+	if TLSMinVersionIndex > TLSMaxVersionIndex &&
+		TLSMaxVersionIndex != -1 &&
+		TLSMinVersionIndex != -1 {
+		return nil, fmt.Errorf("cipher suite flag min version is greater than max version")
```
tuminoid, line 248: It is not cipher suite, but TLS version.

main.go
```diff
@@ -96,13 +108,38 @@ func main() {
 
 	flag.IntVar(&restConfigBurst, "kube-api-burst", 30,
 		"Maximum number of queries that should be allowed in one burst from the controller client to the Kubernetes API server. Default 30")
+	flag.StringVar(&tlsOptions.TLSMinVersion, "tls-min-version", "VersionTLS10",
```
tuminoid, line 111: We certainly do not support TLS 1.0. Minimum supported is 1.2

main.go
```diff
@@ -96,13 +108,38 @@ func main() {
...
+		"The minimum TLS version in use by the webhook server.\n"+
+			fmt.Sprintf("Possible values are %s.", strings.Join(cliflag.TLSPossibleVersions(), ", ")),
+	)
+
+	flag.StringVar(&tlsOptions.TLSMaxVersion, "tls-max-version", "VersionTLS12",
```
tuminoid, line 116: Maximum should be TLS 1.3 as well to allow server/client to choose between 1.2 and 1.3 by default.

main.go
```diff
@@ -185,3 +223,85 @@ func setupWebhooks(mgr ctrl.Manager) {
...
+
+	if options.TLSCipherSuites != "" {
+		tlsCipherSuites := strings.Split(options.TLSCipherSuites, ",")
+		suites, err := cliflag.TLSCipherSuites(tlsCipherSuites)
+		if err != nil {
+			return nil, err
+		}
+
+		if options.TLSMinVersion == "VersionTLS13" {
+			tls13Suites := []string{"TLS_AES_128_GCM_SHA256", "TLS_AES_256_GCM_SHA384", "TLS_CHACHA20_POLY1305_SHA256"}
```
tuminoid, line 275: I don't think it is a good idea to hard code TLS 1.3 suites here. Rather implement the check so that if TLS min and max version is 1.3, then do not allow cipher suites flag to be used at all.

main_test.go
```diff
@@ -0,0 +1,63 @@
...
+		_, err := GetTLSOptionOverrideFuncs(tlsMockOptions)
+		g.Expect(err.Error()).To(Equal("cipher suite TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256 is insecure please update"))
+	})
+}
+
+func TestTLSMinAndMaxVersion(t *testing.T) {
+	t.Run("should fail if tls min version is greater than max version.", func(t *testing.T) {
+		g := NewWithT(t)
+		tlsMockOptions := TLSOptions{
+			TLSMaxVersion: "VersionTLS11",
```
tuminoid, line 42: As said earlier, TLS 1.0 and TLS 1.1 are deprecated and should not be supported at all.

main_test.go
```diff
@@ -0,0 +1,63 @@
...
+			TLSMinVersion: "VersionTLS13",
+		}
+		_, err := GetTLSOptionOverrideFuncs(tlsMockOptions)
+		g.Expect(err.Error()).To(Equal("cipher suite flag min version is greater than max version"))
```
tuminoid, line 46: cipher suite -> tls version

main.go
```diff
@@ -185,3 +223,85 @@ func setupWebhooks(mgr ctrl.Manager) {
...
+				}
+			}
+		}
+		insecureCipherValues := cliflag.InsecureTLSCipherNames()
+		for _, cipher := range tlsCipherSuites {
+			_, containsInsecureCipher := contains(insecureCipherValues, cipher)
+			if containsInsecureCipher {
+				return nil, fmt.Errorf("cipher suite %s is insecure please update", cipher)
+			}
+		}
```
lentzi90, line 290: I think it would be nicer to get rid of the custom `contains` function and do it like here
https://github.com/kubernetes/kubernetes/blob/93ec57734fbb97b33d6060af834a1dfe045fd914/staging/src/k8s.io/apiserver/pkg/server/secure_serving.go#L63-L73
I would also say that we should not error out because the cipher suite is insecure, we should just log a warning (as in the link).

tuminoid:
/retitle :sparkles: Add TLS configuration flags for IPAM

main.go
```diff
@@ -96,13 +109,38 @@ func main() {
...
 
 	if err := logsv1.ValidateAndApply(logOptions, nil); err != nil {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
+	tlsOptionOverrides, err := GetTLSOptionOverrideFuncs(tlsOptions)
+	if err != nil {
+		setupLog.Error(err, "unable to add TLS settings to the webhook server")
+		os.Exit(1)
+	}
```
lentzi90, line 143: This should go below the `ctrl.SetLogger(klog.Background())` so we can use the logger.

main.go
```diff
@@ -185,3 +224,77 @@ func setupWebhooks(mgr ctrl.Manager) {
...
+	tlsMaxVersion, err := cliflag.TLSVersion(options.TLSMaxVersion)
+	if err != nil {
+		return nil, err
+	}
+	tlsOptions = append(tlsOptions, func(cfg *tls.Config) {
+		cfg.MaxVersion = tlsMaxVersion
+	})
+	// Cipher suites should not be set if empty.
+	if options.TLSMinVersion == "VersionTLS13" && options.TLSMaxVersion == "VersionTLS13" {
+		log.Print("cipher suite flag set to null for TLS version 1.3")
```
lentzi90, line 265: Please use the same logger as elsewhere in the file, i.e. `setupLog`:
```suggestion
		setupLog.Info("warning: cipher suite flag set to null for TLS version 1.3")
```

main.go
```diff
@@ -185,3 +224,77 @@ func setupWebhooks(mgr ctrl.Manager) {
...
+		suites, err := cliflag.TLSCipherSuites(tlsCipherSuites)
+		if err != nil {
+			return nil, err
+		}
+
+		insecureCipherValues := cliflag.InsecureTLSCipherNames()
+		for _, cipher := range tlsCipherSuites {
+			for _, insecureCipherName := range insecureCipherValues {
+				if insecureCipherName == cipher {
+					log.Printf("use of insecure cipher '%s' detected.", cipher)
```
lentzi90, line 280:
```suggestion
					setupLog.Info("warning: use of insecure cipher '%s' detected.", cipher)
```

as20203: @lentzi90  do you have any idea how can i fetch these logs inside a buffer in my test files so that I can test them. I was able to do that with the logger library but trying the same here doesn't seem to capture the logs.

lentzi90: Oh, you are checking the log output. I'm not sure if that is possible in this way. Maybe this could be used?
https://pkg.go.dev/k8s.io/klog/v2/ktesting

Anyway, I think it may be better to not check the log output and just see that the function returns the expected configuration

as20203: I was able to fix this logging issue without using ktesting since this library is experimental. I was calling the logger using an incorrect way in the tests. Actually there is a test which detect if insecure cipher are passed so for that checking the output doesn't seem the right way since the output will contain those values. Other than that the output of this function under test  []func(*tls.Config), error is an array of functions with memory pointers which can't be tested. Returning the configuration from this function in struct form would be redundant since It won't be used anywhere in the original code. Going through different articles such as 
https://goutamtadi.medium.com/unit-testing-logs-with-logr-logger-in-golang-59cfa07f9178
suggested this way to do this as well.

lentzi90: Nice! This looks like a good solution to me!

main_test.go
```diff
@@ -0,0 +1,81 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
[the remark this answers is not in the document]
tuminoid: Question: Why do we attribute the work to Kubernetes Authors, and not Metal3 Authors or something?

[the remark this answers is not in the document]
as20203: I've updated this to 2023. Other than that I saw that kubernetes authors were attributed in main.go so I used the same here. According to my knowledge since we are using their libraries in our code, maybe that could be the reason for this.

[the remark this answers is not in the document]
kashifest: Legacy staffs , mainly because these were copied from templates from CAPI and CAPI as well as a few other providers which live in  k8s sigs org use the same, I believe we can either remove this or attribute this to correct authors. Have to check with people with knowledge on Licenses if its fine or not.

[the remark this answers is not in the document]
tuminoid: @kashifest to comment after he has asked around.

main.go
```diff
@@ -185,3 +223,77 @@ func setupWebhooks(mgr ctrl.Manager) {
...
+		return nil, err
+	}
+	tlsOptions = append(tlsOptions, func(cfg *tls.Config) {
+		cfg.MaxVersion = tlsMaxVersion
+	})
+	// Cipher suites should not be set if empty.
+	if options.TLSMinVersion == "VersionTLS13" && options.TLSMaxVersion == "VersionTLS13" {
+		setupLog.Info("warning: cipher suite flag set to null for TLS version 1.3")
+		options.TLSCipherSuites = ""
+	}
```
lentzi90, line 266: I would expect warning only if user tried to set the cipher suites.

main.go
```diff
@@ -185,3 +223,77 @@ func setupWebhooks(mgr ctrl.Manager) {
...
+					setupLog.Info(fmt.Sprintf("warning: use of insecure cipher '%s' detected.", cipher))
+				}
+			}
+		}
+		tlsOptions = append(tlsOptions, func(cfg *tls.Config) {
+			cfg.CipherSuites = suites
+		})
+	}
+
+	if options.TLSCertFile != "" && options.TLSPrivateKeyFile != "" {
```
lentzi90, line 288: If only 1 option is specified we should error out I think. Now we just silently ignore it

lentzi90 (commented):
/approve cancel
until comments addressed

main.go
as20203: @lentzi90 are these required parameters? Because I assumed these were optional values and designed test cases around that scenario. If these are required. What should be their values i.e. is there some place where we have these files located? Otherwise if we provide dummy values for these it will also error out in that scenario.

lentzi90: No not required. But if a cert is specified and no key, then that is an error. I'm open to suggestions on how to handle it but I don't think it is ok that the user can set `--tls-cert-file=/my/certificate` and it is silently ignored.

Example:

- `--tls-cert-file` is specified but not `--tls-private-key-file` -> error
- `--tls-private-key-file` is specified but not`--tls-cert-file` -> error
- both flags are specified -> success
- no flags are specified -> success (don't use TLS)

tuminoid: > * no flags are specified -> success (don't use TLS)

This is not something we want to support. Currently when we don't provide any certs or keys, we use cert-manager provided certs. So if no flags are specified, we keep using default certs.

lentzi90: Right. Let's just get rid of these flags. The webhook-cert-dir is enough and it just makes it confusing/complex to have separate flags for cert and key.

main.go
as20203: fixed this and added additional check when cipher suites are provided and version are TLS 1.3

main_test.go
```diff
@@ -0,0 +1,81 @@
+/*
+Copyright 2023 The Kubernetes Authors.
```
lentzi90, line 2:
```suggestion
Copyright 2023 The Metal3 Authors.
```

main.go
```diff
@@ -185,3 +217,69 @@ func setupWebhooks(mgr ctrl.Manager) {
...
+		return nil, err
+	}
+	tlsOptions = append(tlsOptions, func(cfg *tls.Config) {
+		cfg.MaxVersion = tlsMaxVersion
+	})
+	// Cipher suites should not be set if empty.
+	if options.TLSMinVersion == "VersionTLS13" &&
+		options.TLSMaxVersion == "VersionTLS13" &&
+		options.TLSCipherSuites != "" {
+		setupLog.Info("warning: Cipher suites should not be set for TLS version 1.3")
```
lentzi90, line 260: I would clarify here that we are ignoring the ciphers.
```suggestion
		setupLog.Info("warning: Cipher suites should not be set for TLS version 1.3. Ignoring ciphers")
```

lentzi90 (commented):
Better and better, just a couple of nits below

lentzi90 (commented):
/approve

tuminoid:
Building manager and running `./bin/manager --help` says:

```
  -tls-max-version string
    	The maximum TLS version in use by the webhook server.
    	Possible values are VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13. (default "VersionTLS13")
  -tls-min-version string
    	The minimum TLS version in use by the webhook server.
    	Possible values are VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13. (default "VersionTLS12")
```

We are not supporting TLS 1.0 or TLS 1.1.

as20203:
> Building manager and running `./bin/manager --help` says:
> 
> ```
>   -tls-max-version string
>     	The maximum TLS version in use by the webhook server.
>     	Possible values are VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13. (default "VersionTLS13")
>   -tls-min-version string
>     	The minimum TLS version in use by the webhook server.
>     	Possible values are VersionTLS10, VersionTLS11, VersionTLS12, VersionTLS13. (default "VersionTLS12")
> ```
> 
> We are not supporting TLS 1.0 or TLS 1.1.

Do we build with make run. I tested it previously and it was working fine. Now I'm getting some error 

```"
Could not wait for Cache to sync" err="failed to wait for ippool caches to sync: timed out waiting for cache to be synced" controller="ippool" controllerGroup="ipam.metal3.io" controllerKind="IPPool"
```
```"
"controller-runtime/source: if kind is a CRD, it should be installed before calling Start" err="no matches for kind \"IPPool\" in version \"ipam.metal3.io/v1alpha1\"" kind="IPPool.ipam.metal3.io"
I wanted to test my changes. Since now i've updated the comment.
```

tuminoid:
Please update the git commit message as it still mentions flags were not offering.

Otherwise, I've tested this and it works. Updated description with more information to match functionality.

tuminoid (approved):
/lgtm

Verified that it works. Controller runtime uplift which affects the `TLSInfo` location will be happening separately.

PR #253, opened 2023-06-07
🌱 Add dependabot workflow

kashifest:
This PR adds a workflow job description for dependabot.

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,0 +1,48 @@
+name: dependabot
+
+on:
+  pull_request:
+    branches:
+      - dependabot/**
+  push:
+    branches:
+      - dependabot/**
+  workflow_dispatch:
+
+permissions:
+  contents: write # Allow to update the PR.
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code into the Go module directory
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,48 @@
...
+          ${{ runner.os }}-go-
+    - name: Update all modules
+      run: make modules
+    - name: Update generated code
+      run: make generate
+    - uses: EndBug/add-and-commit@1bad3abcf0d6ec49a5857d124b0bfb52dc7bb081 # tag=v9.1.3
+      name: Commit changes
+      with:
+        author_name: dependabot[bot]
+        author_email: 49699333+dependabot[bot]@users.noreply.github.com
```
kashifest, line 46: need to verify this

kashifest: verified, not an issue, different providers are using the same

lentzi90 (commented):
/approve

lentzi90:
> > This is copy from cluster-api, and the point of the job is to verify build whenever dependabot is proposing bumps via the dependency update workflow?
> 
> Yes

I would say it is not to verify build, that would anyway be checked in the PR it creates. This is to make sure all go modules and other "generated" files are up to date in the PR. Otherwise the checks would fail.

PR #266, opened 2023-06-15
🌱 Include all gomods in different directories for Dependabot

kashifest:
Dependabot is currently not updating gomodules in api/ and hack/tools dir. This PR updates it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,8 +22,30 @@ updates:
       # Ignore k8s and its transitives modules as they are upgraded manually
       # together with controller-runtime.
       - dependency-name: "k8s.io/*"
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/approve

tuminoid:
/hold

Why are we not including `test` as well?

lentzi90:
> Why are we not including test as well?

It doesn't have any `go.mod` so it would not make sense

tuminoid:
> > Why are we not including test as well?
> 
> It doesn't have any `go.mod` so it would not make sense

Duh... it is only for fuzzing stuff in IPAM. Lol me.

/unhold

PR #276, opened 2023-06-22
:sparkles: Added TLS configuration flags for IPAM

as20203:
Added tls configuration flags to IPAM

This enables flags --tls-min-version and --tls-max-version flags, which can be used to limit TLS versions server accepts. Choices are TLS12 and TLS13, so you can force TLS 1.2 only, TLS 1.3 only, or (default) TLS 1.2 and TLS 1.3 is negotiated between server and client.

This also disables TLS 1.0 and TLS 1.1 as they're insecure and obsolete and not supported by Go TLS library.

Manual backport of https://github.com/metal3-io/ip-address-manager/pull/248

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,9 +18,11 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	ipamv1 "github.com/metal3-io/ip-address-manager/api/v1alpha1"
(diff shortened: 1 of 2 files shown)


tuminoid:
/retitle :sparkles: Added TLS configuration flags for IPAM

tuminoid (commented):
/lgtm

lentzi90 (commented):
/approve

PR #277, opened 2023-06-28
:bug: Disable TLS 1.0 and TLS 1.1

as20203:
We found out that TLS 1.0 and TLS 1.1 are enabled despite being insecure. These changes are backported to release-1.3 branch, but without the configuration flags. Min version is set to TLS12 and Max version set to TLS13, with no config to change those.

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"flag"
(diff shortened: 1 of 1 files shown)


tuminoid (commented):
/lgtm

lentzi90 (commented):
/approve

PR #280, opened 2023-07-13
WIP: Bump capi v1.5.0-beta.1 and golang version to v1.20

Sunnatillo:
Bumps capi to v1.5.0-beta.1 and golang version to v1.20.


ipam/ippool_manager_test.go
```diff
@@ -502,53 +511,6 @@ var _ = Describe("IPPool manager", func() {
...
-						},
-						Address: ipamv1.IPAddressStr("192.168.1.11"),
-						Gateway: (*ipamv1.IPAddressStr)(pointer.String("192.168.0.1")),
-						Prefix:  24,
-					},
-				},
-			},
-			expectedAllocations:   map[string]ipamv1.IPAddressStr{},
-			expectedNbAllocations: 0,
-		}),
```
kashifest, line 551: what is this change related to?

kashifest (changes requested):
You need to bump controller-runtime in [api/go.mod](https://github.com/metal3-io/ip-address-manager/blob/main/api/go.mod#L13) as well to v0.15.0

ipam/ippool_manager_test.go
```diff
@@ -384,6 +384,12 @@ var _ = Describe("IPPool manager", func() {
 							Namespace: "myns",
 						},
 					},
+					Status: ipamv1.IPClaimStatus{
+						Address: &corev1.ObjectReference{
+							Name:      "abcpref-192-168-1-11",
+							Namespace: "myns",
+						},
+					},
```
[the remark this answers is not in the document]
Sunnatillo: When IpAddress created, it will update ipClaimStatus firtst and then will proceed with ip address creation. Here somehow we did not include status in IpClaim which is causing an error. Not sure how it was passing before.

config/crd/bases/ipam.metal3.io_ipaddresses.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.4
+    controller-gen.kubebuilder.io/version: v0.12.1
```
kashifest, line 6: CAPI is sitting currently on v0.12.0. Its fine for now since its a patch version but lets wait for CAPI to bump this for future bumps.

hack/tools/go.mod
```diff
@@ -4,8 +4,8 @@ go 1.20
 
 require (
 	github.com/jteeuwen/go-bindata v3.0.7+incompatible
-	k8s.io/code-generator v0.26.1
-	sigs.k8s.io/controller-tools v0.11.4
+	k8s.io/code-generator v0.27.1
```
kashifest, line 7: you have bumped other k8s packages to v0.27.2 in root go.mod and api/go.mod. Can you check if we can do the same for this package.

ipam/ippool_manager_test.go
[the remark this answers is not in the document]
kashifest: Interesting, it shouldn't have passed if thats the case.

kashifest (commented):
Very good job. Few small nits inline. This PR is in pretty good shape.

kashifest:
/retitle Bump CAPI to v1.5.0, controller-runtime to v0.15.0 and controller-tools to v0.12.1

ipam/ippool_manager_test.go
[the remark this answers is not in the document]
Sunnatillo: It was actually due to status handling changes in  [CR v0,15.0](https://github.com/kubernetes-sigs/controller-runtime/pull/2259). Fixed now

kashifest (dismissed):

Sunnatillo:
/hold 
need check the change

ipam/ippool_manager_test.go
```diff
@@ -416,7 +422,7 @@ var _ = Describe("IPPool manager", func() {
 					},
 					Status: ipamv1.IPClaimStatus{
 						Address: &corev1.ObjectReference{
-							Name:      "abcpref-192-168-1-12",
+							Name:      "abcpref-192-168-1-10",
```
smoshiur1237, line 425: Would you please check if this is correct? I think it doesn't need this change.

Sunnatillo: yes, thank you for noticing. Fixed it.

smoshiur1237 (commented):
Thanks for the fix. 
/lgtm

kashifest (approved):

furkatgofurov7:
/retitle ⚠️ Bump CAPI to v1.5.0, controller-runtime to v0.15.0 and controller-tools to v0.12.1

PR #284, opened 2023-07-25
✨  Bump golang to v1.20 and golangci-lint to v1.52.1

kashifest:
We also modify the way we install golangci-lint. Previously we were downloading and installing it through gomodules, however newer versions are not woorking properly with that installation method. Now, with this patch we will be installing it through the workflow file and the ensure-golangci-lint.sh which downloads the binary from github release. We also fix the lint errors here.


kashifest:
This PR needs https://github.com/metal3-io/project-infra/pull/555 to land first. 
/test-centos-e2e-integration-main 
/test-ubuntu-integration-main

Sunnatillo (commented):
LGTM

smoshiur1237 (approved):
/lgtm

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,33 @@
...
+        id: vars
+        run: echo "go_version=$(make go-version)" >> $GITHUB_OUTPUT
+      - name: Set up Go
+        uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753 # v4.0.1
+        with:
+          go-version: ${{ steps.vars.outputs.go_version }}
+      - name: golangci-lint-${{matrix.working-directory}}
+        uses: golangci/golangci-lint-action@639cd343e1d3b897ff35927a75193d57cfcba299 # v3.6.0
+        with:
+          version: v1.52.1
```
furkatgofurov7, line 32: we could bump it to newer 1.53.1 also, but this is also fine

furkatgofurov7 (approved):

.github/workflows/golangci-lint.yml
kashifest: Yes I will do a followup in IPAM and CAPM3 both. Thanks

PR #314, opened 2023-09-06
 🌱 Remove unused scripts and Makefile phonys

kashifest:
Removing the unused binaries. We can add those back once we need it for ipam e2e.


Makefile
```diff
@@ -85,10 +84,6 @@ unit: ## Run tests
 .PHONY: test  ## Run formatter, linter and tests
 test: generate fmt lint unit
 
-.PHONY: test-integration
-test-integration: ## Run integration tests
-	source ./hack/fetch_ext_bins.sh; fetch_tools; setup_envs; go test -v -tags=integration ./test/integration/...
-
 .PHONY: test-e2e
 test-e2e: ## Run e2e tests
```
kashifest, line 88: unsure about the idea here, this is currently broken as well.

lentzi90: Looks like copy paste. We can remove it.

Makefile
```diff
@@ -266,10 +258,6 @@ set-manifest-pull-policy:
 ## Deploying
 ## --------------------------------------
 
-# Run against the configured Kubernetes cluster in ~/.kube/config
-run: generate fmt vet
-	go run ./main.go
-
```
lentzi90, line 272: This one is working fine so I think we can keep it.

kashifest: humm, how is it working for you? I get the following error

```
E0907 09:05:09.329242   24327 main.go:178] "setup: problem running manager" err="open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory"
exit status 1
```
This deploys the webhook as well, and simply running main.go wont work, which was the reason we removed the run phony from CAPM3 as well.

lentzi90: Well I mean if you put a certificate there it works fine. Maybe it is not so useful, I just thought that there is nothing "broken" with this since it runs the manager as expected.

kashifest: This used to work without deploying certificate when webhook server was not deployed with manager but separately. Now you need to generate and deploy certificates to make this phony to work, I believe tilt deploy would be much easier. I can add the build phonies as we have in CAPM3 though, those may be useful to check if we can build the go modules.

lentzi90 (commented):
/approve

tuminoid (commented):
/lgtm

PR #323, opened 2023-09-21
🐛 fix ensure-golangci-lint digest for Linux/amd64

tuminoid:
The digest for Linux/amd64 for ensure-golangci-lint are incorrect, and also the script fails to unbound variable when checking the SHA.

This fixes "make lint", and "make test", which are no longer utilized as CI tests, but only run locally.

Makes the script consider ARCH as well, which is the root cause for these mismatches.

Later on, I will add project-infra test to actually verify these scripts work.


tuminoid:
/cc @kashifest 
Please test this on Mac locally.

lentzi90 (commented):
/approve

smoshiur1237 (commented):
/lgtm

PR #369, opened 2023-11-01
🌱 Adding PR title verifier

adilGhaffarDev:
Adding PR title verifier. This uses https://github.com/kubernetes-sigs/kubebuilder-release-tools and makes sure that the title has an icon added. This will make the release process easier.

### DIFF
--- a/.github/workflows/pr-verifier.yaml
+++ b/.github/workflows/pr-verifier.yaml
@@ -1,0 +1,19 @@
+name: PR Verifier
+
+on:
+  pull_request_target:
+    types: [opened, edited, reopened, synchronize]
+
+permissions:
+  contents: read
+
+jobs:
+  verify:
+    runs-on: ubuntu-latest
+    name: verify PR contents
+    steps:
+    - name: Verifier action
+      id: verifier
+      uses: kubernetes-sigs/kubebuilder-release-tools@e389278c2051863fb39b8de00159c8d7fac760ea # v0.4.1
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/pr-verifier.yaml
```diff
@@ -0,0 +1,19 @@
+name: PR Verifier
+
+on:
+  pull_request_target:
+    types: [opened, edited, reopened, synchronize]
+
+permissions:
+  checks: write
```
tuminoid, line 8: Does it really need to write? None of our existing workflows have it.

adilGhaffarDev:
cc @tuminoid @kashifest

.github/workflows/pr-verifier.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  verify:
+    runs-on: ubuntu-latest
+    name: verify PR contents
+    steps:
+    - name: Verifier action
+      id: verifier
+      uses: kubernetes-sigs/kubebuilder-release-tools@e389278c2051863fb39b8de00159c8d7fac760ea # tag=v0.4.1
```
tuminoid, line 17:
```suggestion
      uses: kubernetes-sigs/kubebuilder-release-tools@e389278c2051863fb39b8de00159c8d7fac760ea # v0.4.1
```
It should be withotu `tag=` part. Only single workflow in IPAM has `tag=`, which we should correct to be consistent.

.github/workflows/pr-verifier.yaml
adilGhaffarDev: yes, we don't need it, I will remove it.

.github/workflows/pr-verifier.yaml
adilGhaffarDev: fixed

.github/workflows/pr-verifier.yaml
adilGhaffarDev: removed

.github/workflows/pr-verifier.yaml
```diff
@@ -0,0 +1,16 @@
+name: PR Verifier
+
+on:
+  pull_request_target:
+    types: [opened, edited, reopened, synchronize]
+
```
tuminoid, line 6: Now we don't have any permissions, which is also bad. If it would be my guess, we need 
```yaml
permissions:
  contents: read
```
since we're reading the PR title.

tuminoid (commented):
/lgtm

kashifest (commented):
/approve

kashifest:
/hold
Shouldn't we see the workflow running on this PR already?

kashifest:
> /hold Shouldn't we see the workflow running on this PR already?

/hold cancel
Lets merge it and see if it works afterwards

adilGhaffarDev:
> > /hold Shouldn't we see the workflow running on this PR already?
> 
> /hold cancel Lets merge it and see if it works afterwards

I am checking

PR #371, opened 2023-11-02
🌱 test commit

adilGhaffarDev:
test
Fixes #

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -60,3 +60,5 @@ using fake clients from controller-runtime package
 * The tests should be table-driven when there are multiple cases.
 * Table-driven tests should use subtests to improve the readability of the
   results
+
+Test commits.


adilGhaffarDev:
@kashifest it is running and failed: https://github.com/metal3-io/ip-address-manager/actions/runs/6729695242/job/18290998859?pr=371

kashifest:
> @kashifest it is running and failed: https://github.com/metal3-io/ip-address-manager/actions/runs/6729695242/job/18290998859?pr=371

Yes perhaps it requires write permission after all. cc @tuminoid

adilGhaffarDev:
> Yes perhaps it requires write permission after all. cc @tuminoid

maybe, in capi they have added the permission. I will add it, lets see.

kashifest:
> maybe, in capi they have added the permission. I will add it, lets see.

We added it already, its now working, I mean failing on PRs where we dont have correct titles, albeit we have to only see the failure reason on the logs of the workflow Run. Check #370

PR #378, opened 2023-11-15
✨ Prepare metal3-ipam as a CAPI IPAM provider

kashifest:


metadata.yaml
```diff
@@ -0,0 +1,18 @@
...
+  contract: v1beta1
+- major: 1
+  minor: 3
+  contract: v1beta1
+- major: 1
+  minor: 2
+  contract: v1beta1
+- major: 1
+  minor: 1
+  contract: v1beta1
```
kashifest, line 18: @lentzi90 do you think we need to add the older contracts here? I already skipped the v1ax contracts, IMO adding older v1b1 contracts doesn't make much sense, wdyt?

lentzi90: Yes I think we can safely skip them. This looks good!

Rozzii:
@kashifest  how time sensitive it is, would it be fine to rebae this after we have pushed DCO enforcement?

kashifest:
> @kashifest how time sensitive it is, would it be fine to rebae this after we have pushed DCO enforcement?

Yes, it can wait
/hold

kashifest:
This would require more changes, @peppi-lotta will take care of it in a seperate PR

PR #394, opened 2023-12-01
🌱 Add tag triggering for container building wf

mquhuy:
With this PR creation of git tags with pattern `v**` will trigger container image build

### DIFF
--- a/.github/workflows/build-images-action.yml
+++ b/.github/workflows/build-images-action.yml
@@ -4,7 +4,8 @@ on:
     branches:
     - 'main'
     - 'release-*'
-
+    tags:
+    - 'v*'
 permissions: {}
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-images-action.yml
```diff
@@ -4,6 +4,8 @@ on:
     branches:
       - 'main'
```
tuminoid, line 5: Further confusion about using or not using single quotes. Here the main is also quoted, while most other workflow files it is not.

mquhuy: This was added after Kashif commented that it needs a quote, but I later found out that the quote is not needed in GH actions doc.

mquhuy: Thanks @smoshiur1237 . Anw after all I decided to add quotes everywhere for the shake of consistency and to prevent future confusions.

smoshiur1237 (approved):
/lgtm

tuminoid:
> /test-centos-e2e-integration-main /test-ubuntu-integration-main

Doh, not supposed to run them for this one. Ignore.

tuminoid (commented):
/lgtm

kashifest (commented):
/approve

tuminoid:
This is good to unhold now?

.github/workflows/build-images-action.yml
tuminoid, line 1: You're reindenting the file, can you just fix the params?

tuminoid (changes requested):

tuminoid (approved):
/lgtm

PR #400, opened 2023-12-14
🌱 Use k8s apimachinery scheme builder instead of controller-runtime scheme builder

kashifest:
This is recommend for the CAPI providers as mentioned the v1.5 to v1.6 migration guide: https://cluster-api.sigs.k8s.io/developer/providers/migrations/v1.5-to-v1.6#suggested-changes-for-providers.

The original issue was discussed in https://github.com/kubernetes-sigs/cluster-api/issues/9011. The issue is using controller-runtime based scheme builder causes a lot of unnecessary dependancies added in go modules. Compared to that apimachinery was chosen as a better alternative. 

Relevant CAPM3 PR# https://github.com/metal3-io/cluster-api-provider-metal3/pull/1360

### DIFF
--- a/api/v1alpha1/groupversion_info.go
+++ b/api/v1alpha1/groupversion_info.go
@@ -23,23 +23,29 @@ limitations under the License.
 package v1alpha1
 
 import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


lentzi90 (commented):
/approve

smoshiur1237 (commented):
/lgtm

PR #408, opened 2024-01-08
🌱 reduce github action permissions

tuminoid:
`contents: write` is excessive right for dependabot, as we don't allow dependabot to auto-merge anything. It is enough for it to read contents and create pull request.

Move permissions from top-level to job-level as well.

Cleanup actions formatting to be the same for easier reading and maintenance. Add missing version tags.

These is highlighted by the CLOmonitor/OpenSSF Scorecard. This restores score for token permissions to 10/10.

### DIFF
--- a/.github/workflows/build-images-action.yml
+++ b/.github/workflows/build-images-action.yml
@@ -2,27 +2,31 @@ name: build-images-action
 on:
   push:
     branches:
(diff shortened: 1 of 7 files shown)


lentzi90 (commented):
/lgtm

PR #415, opened 2024-01-15
🌱 fix dependabot workflow permissions

tuminoid:
Dependabot itself wouldn't need more than pull-requests: write, but we have added on generated code part which does need to write contents on top of the PR, so restoring the permissions.

Symptoms are seen in #414


tuminoid:
/override test-centos-e2e-integration-main test-ubuntu-integration-main 
Irrelevant.

kashifest (commented):
/approve

smoshiur1237 (commented):
/lgtm

PR #427, opened 2024-02-06
⚠️ Changes to make IPAM deployable with CAPI

peppi-lotta:
These changes allow metal3 ip-address-manager(IPAM) to be deployed with cluster-api (CAPI). Currently deploying IPAM with CAPI assumes the CRDs are in a git release and the release has a metadata.yaml.  IPAM's current release doesn't have a metadata.yaml so this need to be added. Action need to be taken to get IPAM as an ipam-provider for CAPI so that metal3IPAM can be specified when running clusterctl init with the --ipam flag. 

IPAM's ipam-components.yaml has CRDs for ipClaim and ipAddresses which work but IPAM is unable to resolve the ipAddresClaims that CAPI has defined. This is a next step after these changes are
meged to make IPAM resolve CAPI's ipAddressClaims. Check more in [CAPI ipam contract](https://github.com/kubernetes-sigs/cluster-api/pull/10108/commits/390031158f82d31577e7fe28732670971f173f6d)

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -1,9 +1,8 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
-# Adds namespace to all resources. Keep it in capm3-system, as it is a
-# dependency for CAPM3
-namespace: capm3-system
+# Adds namespace to all resources
+namespace: ipam-provider-m3-system
 
 namePrefix: ipam-
 
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -3,6 +3,7 @@ kind: Kustomization
 
 resources:
 - manager.yaml
+- namespace.yaml
 
 generatorOptions:
  disableNameSuffixHash: true
--- a/config/manager/namespace.yaml
+++ b/config/manager/namespace.yaml
@@ -1,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: system
--- a/examples/generate.sh
+++ b/examples/generate.sh
@@ -23,7 +23,7 @@ OUTPUT_DIR=${OUTPUT_DIR:-${SOURCE_DIR}/_out}
 
 # Cluster.
 export CLUSTER_NAME="${CLUSTER_NAME:-test1}"
-export NAMESPACE="${NAMESPACE:-capm3-system}"
+export NAMESPACE="${NAMESPACE:-ipam-provider-m3-system}"
 
 # Outputs.
 COMPONENTS_CERT_MANAGER_GENERATED_FILE=${OUTPUT_DIR}/cert-manager.yaml
--- a/examples/provider-components/manager_tolerations_patch.yaml
+++ b/examples/provider-components/manager_tolerations_patch.yaml
@@ -3,7 +3,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: ipam-controller-manager
-  namespace: capm3-system
+  namespace: ipam-provider-m3-system
 spec:
   template:
     spec:


config/default/kustomization.yaml
```diff
@@ -1,9 +1,8 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
-# Adds namespace to all resources. Keep it in capm3-system, as it is a
-# dependency for CAPM3
-namespace: capm3-system
+# Adds namespace to all resources
+namespace: ipam-provider-m3-system
```
lentzi90, line 5: This is a bit of a mouth full. Could we simplify it a bit?
I suggest `ipam-metal3-system`. For reference the CAPI in-cluster IPAM has the namespace `capi-ipam-in-cluster-system` so following that we would get `capi-ipam-metal3-system`, but since we can use this also without CAPI, I think we can drop the `capi-` prefix.
```suggestion
namespace: ipam-metal3-system
```

config/default/kustomization.yaml
```diff
@@ -1,9 +1,8 @@
...
 
 namePrefix: ipam-
```
lentzi90, line 7: This prefix must match so that it + the namespace `name` (=system) becomes the `namespace` above.
```suggestion
namePrefix: ipam-metal3-
```

kashifest:
@furkatgofurov7 @schrej can we get some reviews from you folks on this ?

config/manager/kustomization.yaml
```diff
@@ -3,6 +3,7 @@ kind: Kustomization
 
 resources:
 - manager.yaml
+- namespace.yaml
```
furkatgofurov7, line 6: nit: we could directly place ns inside the manager.yaml same way as in https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/blob/main/config/manager/manager.yaml#L1-L6

config/default/kustomization.yaml
furkatgofurov7: +1

furkatgofurov7 (commented):
AFAIK there is no way to make it work without CAPM3 changes, can we verify both changes in CI (should be possible to run CI from CAPM3 pointing to the branch of current change)

config/default/kustomization.yaml
schrej: for the in-cluster provider I went with `capi-ipam-in-cluster-`, so it could be `capi-ipam-metal3-` to stay consistent here. I'd argue that it might help to be consistent, so it's easier to navigate for people that are new to CAPI, but it's entirely up to you.

lentzi90:
@peppi-lotta can you update the description? The part about "in-cluster directory" was a misunderstanding if I recall correctly

kashifest:
> Imho we should avoid deploying custom claims and addresses when using clusterctl to install metal3 ipam. If the plan is to support the CAPI variants of those resources anyway, deploying the custom resources is probably superfluous, and will only lead to confusion (and annoying autocompletion). As long as the official resources aren't supported, I'd also not accept the addition of the metal3 ipam to the official list of providers integrated in clusterctl, as it's not compatible with other providers. Not sure if the CAPI team has the same opinion here.

I want to understand this part a bit more. I can agree to the point that metal3 ipam needs to handle the CAPI variant of the IPAM resources to be accepted as a CAPI IPAM provider. But on the point of deploying metal3 specific resources and CRDs, it should remain as it is now since if a user want to deploy a metal3 IPAM then the user is aware that metal3 IPAM allows you to have this custom claims on top and offers you some additional workflow. So I dont see any harm deploying those. Otherwise if someone doesnt want metal3 IPAM but only CAPI specific variants then the current IPAM provider should be enough.

Rozzii:
What is the current state of this ticket ? @peppi-lotta  could you please summarize it here? The PR went silent in February.

peppi-lotta:
These changes along https://github.com/metal3-io/cluster-api-provider-metal3/pull/1442  help make it possible to deploy this ip-address-manager with clusterctl. This has been on hold while I've been researching the possibility to switch to use in-cluster-ipam in CAPM3. This is currently not possible so I'll continue to make this ip-address-manager an IPAM-provider for CAPI. I'll hopefully be able to continue with this soon.

PR #431, opened 2024-02-22
🌱 Improve ipv6 address in bound check performance in ippool webhook

### LINKED ISSUE #430: ipppol webhook times out on large ipv6 size when adding a preallocation

1. Create an ippool CR with an ipv6 subnet of a large prefix:
`apiVersion: ipam.metal3.io/v1alpha1
kind: IPPool
metadata:
  labels:
    cluster.x-k8s.io/cluster-name: ipam
    cluster.x-k8s.io/provider: infrastructure-metal3
  name: pool4
  namespace: capm3-system
spec:
  clusterName: ipam
  gateway: fda0:d59c:da0a:624::1
  namePrefix: pool4
  pools:
  - gateway: fda0:d59c:da0a:624::1
    start: fda0:d59c:da0a:624::8
    subnet: fda0:d59c:da0a:624::/96
  prefix: 96
`
2. Update the CR with a preAllocation claim that declares an ip address at the end of the cidr range:
`apiVersion: ipam.metal3.io/v1alpha1
kind: IPPool
metadata:
  labels:
    cluster.x-k8s.io/cluster-name: ipam
    cluster.x-k8s.io/provider: infrastructure-metal3
  name: pool4
  namespace: capm3-system
spec:
  clusterName: ipam
  gateway: fda0:d59c:da0a:624::1
  namePrefix: pool4
  pools:
  - subnet: fdad:d59d:da0a:624::/126
  - gateway: fda0:d59c:da0a:624::1
    start: fda0:d59c:da0a:624::8
    subnet: fda0:d59c:da0a:624::/96
  preAllocations:
    claim1: fda0:d59c:da0a:624::ffff:ffff
  prefix: 96
`

kubectl edit cmd will return the following error after timeout of 10 seconds:
`error: ippools.ipam.metal3.io "pool4" could not be patched: Internal error occurred: failed calling webhook "validation.ippool.ipam.metal3.io": failed to call webhook: Post "https://ipam-webhook-service.capm3-system.svc:443/validate-ipam-metal3-io-v1alpha1-ippool?timeout=10s": context deadline exceeded
`

The same would occur if the preallocation claim is out of bound of the large ip pool subnet.

CR update should complete without an error

**Environment:**

- Cluster-api version: 1.5.2
- CAPM3 version:  1.5.1
- IPAM version: 1.5.1
- Minikube version: n/a
- environment (metal3-dev-env or other):
- Kubernetes version: (use `kubectl version`): 1.28.3

/kind bug

jgu17:
The current mechanism of checking if an ip address is within the bound of a given ippool loops through the domain of all ip addresses in the ip pool one by one.  A subnet in ipv6 can easily be a very large cidr that involves millions or billions of addresses. The validation webhook could time out (default 10 seconds) before the check is complete, if the ip address being checked is at the end of the cidr range or even out of the range at all.

Fixes #430

### DIFF
--- a/api/v1alpha1/ippool_webhook.go
+++ b/api/v1alpha1/ippool_webhook.go
@@ -14,6 +14,8 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"net"
+	"net/netip"
 	"reflect"
 
 	"github.com/pkg/errors"
@@ -113,25 +115,49 @@ func (c *IPPool) checkPoolBonds(old *IPPool) ([]IPAddressStr, []IPAddressStr) {
 }
 
 func (c *IPPool) isAddressInBonds(address IPAddressStr) bool {
-	inBonds := false
-	for _, pool := range c.Spec.Pools {
-		if inBonds {
-			break
-		}
-		index := 0
-		for !inBonds {
-			allocatedAddress, err := GetIPAddress(pool, index)
-			if err != nil {
-				break
-			}
-			index++
-			if allocatedAddress == address {
-				inBonds = true
-				break
-			}
-		}
+	ip, err := netip.ParseAddr(string(address))
+	if err != nil {
+		return false
 	}
-	return inBonds
+
+	for _, pool := range c.Spec.Pools {
+		if pool.Start != nil {
+			startIP, err := netip.ParseAddr(string(*pool.Start))
+			if err != nil {
+				// skip this invalid pool, as the validation error should be caught somewhere else
+				continue
+			}
+			if startIP.Compare(ip) > 0 {
+				continue
+			}
+		}
+
+		if pool.End != nil {
+			endIP, err := netip.ParseAddr(string(*pool.End))
+			if err != nil {
+				// skip this invalid pool, as the validation error should be caught somewhere else
+				continue
+			}
+			if endIP.Compare(ip) < 0 {
+				continue
+			}
+		}
+
+		if pool.Subnet != nil {
+			_, subnet, err := net.ParseCIDR(string(*pool.Subnet))
+			if err != nil {
+				// skip this invalid pool, as the validation error should be caught somewhere else
+				continue
+			}
+			if !subnet.Contains(net.ParseIP(ip.String())) {
+				continue
+			}
+		}
+
+		return true
+	}
+
+	return false
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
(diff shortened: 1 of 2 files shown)


kashifest (commented):
Thanks a lot for the fix. This looks good to me overall. One minor comment inline. I will wait for others to have a look before a put a label here.

lentzi90 (commented):
/approve
Thanks for this!

api/v1alpha1/ippool_webhook.go
```diff
@@ -112,26 +114,50 @@ func (c *IPPool) checkPoolBonds(old *IPPool) ([]IPAddressStr, []IPAddressStr) {
...
-			if allocatedAddress == address {
-				inBonds = true
-				break
+			if endIP.Compare(ip) < 0 {
+				continue
 			}
 		}
+
+		if pool.Subnet != nil {
+			_, subnet, err := net.ParseCIDR(string(*pool.Subnet))
```
kashifest, line 147: looks like my comment didn't land. Not a fan of empty vars `_`, we are constantly updating the linters and this might be an issue in near future.

lentzi90: I think the linters would complain if we put something other than `_`? This is a nice way of saying we do not use that variable.

kashifest: Actually yes, its the other way around, sorry for mixing up

api/v1alpha1/ippool_webhook.go
```diff
@@ -96,14 +98,14 @@ func (c *IPPool) checkPoolBonds(old *IPPool) ([]IPAddressStr, []IPAddressStr) {
 	allocationOutOfBonds := []IPAddressStr{}
 	inUseOutOfBonds := []IPAddressStr{}
 	for _, address := range c.Spec.PreAllocations {
-		inBonds := c.isAddressInBonds(address)
+		inBonds := c.isAddressInBounds(address)
```
furkatgofurov7, line 101: I think the original term (Bond) was correct one that we should use here instead of Bound

furkatgofurov7 (commented):
Changes look good to me, it would be even nicer if we could  add a simple test case for `isAddressInBonds` (i.e address in/out of bonds).

api/v1alpha1/ippool_webhook.go
jgu17: I was not very sure what bond meant here, I thought it was trying to check if the address is in the bounds of the pool.  I switched back to the original wording as suggested.

jgu17: There was a positive and negative test cases in the webhook test class, using the start and end in the spec. I just added another pair for using Subnet in the pool spec.

furkatgofurov7: Thank you!

jgu17:
thanks everyone for getting this fix merged!

PR #443, opened 2024-03-04
:seedling: Add make unit-cover like in BMO

### LINKED ISSUE #440: Add unit-cover make target
This issue is reserved for the KubeCon EU 2024 contrib-fest. Please do not work on it before 2024-03-20.

Add a make target for running the unit tests and checking the test coverage. There is a target like this in [BMO](https://github.com/metal3-io/baremetal-operator/blob/b5c4120dd294e94cd63fede91321f38623293ec7/Makefile#L101-L108) that you can use for inspiration.
The goal is to be able to run `make unit-cover`. This should run all unit tests and print test coverage results.

lekaf974:

Fixes #440

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,4 +33,3 @@ releasenotes
 *~
 *.tmp
 .DS_Store
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ releasenotes
 *~
 *.tmp
 .DS_Store
+.idea/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -33,3 +33,4 @@ releasenotes
 *~
 *.tmp
 .DS_Store
+.idea/
```
tuminoid, line 36: Line feed missing.

tuminoid (commented):
/ok-to-test

.gitignore
lekaf974: Removed the row from the file since it was more related to my setup I configured it on my side instead

lentzi90 (commented):
/approve
Thanks!

lekaf974:
Took a look on the test that failed and seems not to be related to those changes

tuminoid (commented):
/lgtm

We can override the ubuntu integration test if it fails again.
Thanks for the contribution!

PR #445, opened 2024-03-04
🌱 fix build image action for release-1.6

tuminoid:
Fix build image action for release-1.6 and make it consistent with the main.

### DIFF
--- a/.github/workflows/build-images-action.yml
+++ b/.github/workflows/build-images-action.yml
@@ -1,28 +1,34 @@
 name: build-images-action
+
 on:
   push:
     branches:
-      - 'main'
-      - 'release-*'
+    - 'main'
+    - 'release-*'
+    tags:
+    - 'v*'
+
 permissions: {}
+
 jobs:
   build:
     name: Build container images
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


tuminoid:
/hold
Until we have verified with IPAM release that it fixes the release tags in release branches as well.

smoshiur1237 (commented):
/lgtm

.github/workflows/build-images-action.yml
```diff
@@ -1,28 +1,34 @@
...
+    tags:
+    - 'v*'
+
 permissions: {}
+
 jobs:
   build:
     name: Build container images
     runs-on: ubuntu-latest
-    if: github.repository == 'metal3-io/ip-address-manager'
```
kashifest, line 12: why are we removing this? It will run the actions on the Nordix/ fork and other forks as well, do we need that

kashifest: ah ignore!!!

kashifest (commented):
/approve

.github/workflows/build-images-action.yml
tuminoid: It is just moved lower, so all the build action files are formatted the same.

PR #449, opened 2024-03-04
🌱 Improve ipv6 address in bound check performance in ippool webhook

jgu17:
What this PR does / why we need it:

Cherry-pick form main to get the fix into release 1.6

The current mechanism of checking if an ip address is within the bound of a given ippool loops through the domain of all ip addresses in the ip pool one by one. A subnet in ipv6 can easily be a very large cidr that involves millions or billions of addresses. The validation webhook could time out (default 10 seconds) before the check is complete, if the ip address being checked is at the end of the cidr range or even out of the range at all.

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes https://github.com/metal3-io/ip-address-manager/issues/430

### DIFF
--- a/api/v1alpha1/ippool_webhook.go
+++ b/api/v1alpha1/ippool_webhook.go
@@ -14,6 +14,8 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"net"
+	"net/netip"
 	"reflect"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


furkatgofurov7 (approved):
This backport can be accepted based on [backport policy](https://github.com/metal3-io/ip-address-manager/blob/main/CONTRIBUTING.md#breaking-changes) of the repo even though it is in `api/` folder (without actually touching any CRD or relevant but just webhooks)

/approve

lentzi90 (commented):
/lgtm

PR #485, opened 2024-04-05
🌱 add workflow to approve actions on PR change

tuminoid:
If external contributor adds a PR, and a member /ok-to-test it, GH actions do not run until approved by an approver. This approval needs to be given if the PR content changes, so it inconveniences the contributor but also the approver.

This PR approves workflows, if ok-to-test label is present.

Adapted from CAPI workflow:
https://github.com/kubernetes-sigs/cluster-api/blob/main/.github/workflows/pr-gh-workflow-approve.yaml

### DIFF
--- a/.github/workflows/pr-gh-workflow-approve.yaml
+++ b/.github/workflows/pr-gh-workflow-approve.yaml
@@ -1,0 +1,42 @@
+# adapted from github.com/kubernetes-sigs/cluster-api/.github/workflows/pr-gh-workflow-approve.yaml
+# this workflow approves workflows if the PR has /ok-to-test
+# related Prow feature request https://github.com/kubernetes/test-infra/issues/25210
+
+name: Approve GH Workflows
+
+on:
+  pull_request_target:
+    types: [opened, edited, reopened, synchronize]
+
+jobs:
+  approve:
+    name: Approve on ok-to-test
+    runs-on: ubuntu-latest
+
+    permissions:
+      actions: write
+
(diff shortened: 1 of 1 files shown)


tuminoid:
Not sure how to test it before merging, but given that it is from CAPI and its working for them, I'd say it should work the same for us.

We can test it here in IPAM first, then take it into CAPM3 and BMO as well (+standalone operator too)

tuminoid:
/cc @lentzi90 @kashifest @elfosardo @dtantsur 
Would you be OK with this?

Sunnatillo (commented):
LGTM

kashifest (commented):
/approve

lentzi90 (commented):
/lgtm

tuminoid:
/hold
To get more opinions. I don't expect objections tho.

dtantsur (approved):

tuminoid:
/override test-centos-e2e-integration-main  test-ubuntu-integration-main

tuminoid:
Tested in PR https://github.com/metal3-io/ip-address-manager/pull/486

Before `/ok-to-test`:

image

After `/ok-to-test`:

image

NOTE, the initial approve needs to be given by the admin. It will not be given by the workflow. The workflow only re-approves once approved workflow, if /ok-to-test is present when PR is updated. This is what we want, so test passed.

tuminoid:
Updated the explanation with case 1 and case 2 just to be super clear what this allows.

PR #507, opened 2024-04-26
Fix multi-architecture images build

haijianyang:
`docker-build-all` cannot build multi-architecture images (https://quay.io/repository/metal3-io/ip-address-manager?tab=tags).

Reference CAPI multi-architecture images build: https://github.com/kubernetes-sigs/cluster-api/blob/main/Dockerfile

Original images:
image

New images:
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,9 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Build architecture
+ARG ARCH
+
 # Support FROM override
 ARG BUILD_IMAGE=docker.io/golang:1.21.9@sha256:7d0dcbe5807b1ad7272a598fbf9d7af15b5e2bed4fd6c4c2b5b3684df0b317dd
-ARG BASE_IMAGE=gcr.io/distroless/static:nonroot@sha256:9ecc53c269509f63c69a266168e4a687c7eb8c0cfd753bd8bfcaa4f58a90876f
+ARG BASE_IMAGE=gcr.io/distroless/static:nonroot-${ARCH}
 
 # Build the manager binary on golang image
 FROM $BUILD_IMAGE as builder


Dockerfile
```diff
@@ -12,9 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Build architecture
+ARG ARCH
+
 # Support FROM override
 ARG BUILD_IMAGE=docker.io/golang:1.21.9@sha256:7d0dcbe5807b1ad7272a598fbf9d7af15b5e2bed4fd6c4c2b5b3684df0b317dd
-ARG BASE_IMAGE=gcr.io/distroless/static:nonroot@sha256:9ecc53c269509f63c69a266168e4a687c7eb8c0cfd753bd8bfcaa4f58a90876f
+ARG BASE_IMAGE=gcr.io/distroless/static:nonroot-${ARCH}
```
tuminoid, line 20: We need to have this pinned, so at a minimum the default needs to stay as it was. If its configurable via `ARCH` on top, thats fine.

tuminoid (commented):
Pinning needs to be addressed.

Also please follow the PR template and add contribution type emoji to title, sign off DCO etc.

/ok-to-test

lentzi90:
Fixes #517.
Could you add overrides for BASE_IMAGE in the Makefile instead? One for each ARCH.

lentzi90:
/hold
I don't think this is needed. The BASE_IMAGE that we use is actually pointing to a multi-arch manifest. The issue is that we simply do not build and publish images for other architectures. I have confirmed this by building an arm64 image on a raspberry pi. If the base image was the issue, that would not have worked.

However, I think you may have run into another issue. We have hard-coded the ARCH to be amd64. That is fixed in https://github.com/metal3-io/ip-address-manager/pull/516.

mboukhalfa:
/close 
Seems not needed as explained above but feel free to reopen if you think still needed

PR #516, opened 2024-05-02
🌱 Set ARCH based on go env

### LINKED ISSUE #517: Support building for multiple architectures
User Story

As a developer/user/operator I would like to use pre-built IPAM images on multiple architectures because it is convenient (compared to building everything yourself).

Detailed Description
We currently only publish amd64 container images, but there are other popular architectures. Go has good support for building for other architectures also, so there is no need for special hardware for testing this.
I suggest we look at how CAPI/CAPM3 has structured the Makefile with a docker-build-all target for building all supported architectures. This should be simple to implement for us as well.
~~Since we do sha pinning in the Dockerfile, we will also need to [override these](https://github.com/metal3-io/ip-address-manager/blob/0d1fd6d7e369492d062de230a9557e77b53adcf0/Dockerfile#L16-L17).~~ This is not needed. The BASE_IMAGE is already pointing to a multi-arch manifest.

Once we can build these images, it should be fairly trivial to also make CI build and publish them together with a multi-arch manifest for easy consumption.

Anything else you would like to add:

/kind feature

lentzi90:

We had ARCH hard-coded to amd64. This commit changes it so that we take it from go env instead. That way we should automatically get the proper architechture on different devices.

Fixes #517

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ PROD_REGISTRY := quay.io/metal3-io
 IMAGE_NAME ?= ip-address-manager
 CONTROLLER_IMG ?= $(REGISTRY)/$(IMAGE_NAME)
 TAG ?= v1alpha1
-ARCH ?= amd64
+ARCH ?= $(shell go env GOARCH)
 ALL_ARCH = amd64 arm arm64 ppc64le s390x
 
 # Allow overriding manifest generation destination directory


smoshiur1237 (approved):
/lgtm

Sunnatillo (commented):
LGTM

kashifest (commented):
/approve

PR #539, opened 2024-05-20
🌱 Optimize the dependabot configuration and add release branch configurations.

kashifest:
This PR add the following:

- Adds newly introduced multi-directories feature in dependabot. With this hopefully all the go modules would be bumped in one PR for a dependancy. Also this PR optimizes the dependabot configs a bit by adding a group for k8s.io dependancy bump and allowing patch versions for the ignore rules.
- Also add release branch configurations for dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,42 +3,33 @@
 
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
-    directory: "/" # Location of package manifests
-    schedule:
-      interval: "monthly"
-    commit-message:
-      prefix: ":seedling:"
+- package-ecosystem: "github-actions"
+  directory: "/" # Location of package manifests
+  schedule:
+    interval: "monthly"
+  commit-message:
+    prefix: ":seedling:"
     # Go
-  - package-ecosystem: "gomod"
-    directory: "/"
-    schedule:
-      interval: "weekly"
-    ignore:
-      # Ignore controller-runtime as its upgraded manually.
-      - dependency-name: "sigs.k8s.io/controller-runtime"
-      # Ignore k8s and its transitives modules as they are upgraded manually
-      # together with controller-runtime.
-      - dependency-name: "k8s.io/*"
-    commit-message:
-      prefix: ":seedling:"
-  - package-ecosystem: "gomod"
-    directory: "/api"
-    schedule:
-      interval: "weekly"
-    ignore:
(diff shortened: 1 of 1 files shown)


tuminoid:
/hold
I don't think we're 100% done with testing in CAPM3 yet with this config. Holding so we don't need to be backporting fixes, but can port a working solution in one go.

kashifest:
> /hold I don't think we're 100% done with testing in CAPM3 yet with this config. Holding so we don't need to be backporting fixes, but can port a working solution in one go.

This is adding a different config, namely the multi directory, so itself this would be tested here, which test you plan to do with CAPM3 ?

tuminoid:
> > /hold I don't think we're 100% done with testing in CAPM3 yet with this config. Holding so we don't need to be backporting fixes, but can port a working solution in one go.
> 
> This is adding a different config, namely the multi directory, so itself this would be tested here, which test you plan to do with CAPM3 ?

I was thinking doing the whole configuration change first in one repo and then bringing in the new config with both branches and directories to others. If we do dirs in one and branches in another, we need to cross-port the missing parts and have two repos where to debug the full config.

tuminoid:
> > I was thinking doing the whole configuration change first in one repo and then bringing in the new config with both branches and directories to others. If we do dirs in one and branches in another, we need to cross-port the missing parts and have two repos where to debug the full config.
> 
> sounds reasonable, although this multi dir thingi is reported as a bit buggy by others, I intended to introduce it in IPAM where we expect less traffic/impact. And the branch config is mostly working as expected in CAPM3.
> 
> Let me know if you think, adding multi-dir in CAPM3 is still a better way to handle this.

Maybe we can wait a bit to see CAPM3 branches config working fine, bring that to IPAM and then add directories on top. This way the config wouldn't be cross-ported and we'd get to test the beta directories config in IPAM first?

kashifest:
sure lets do that. This is anyways ready whenever you feel CAPM3 is tested and working properly.

kashifest:
@tuminoid PTAL, this is now ready to go in

tuminoid (commented):
/approve

This should have IPAM up to date with latest dependabot features and also add the release branch configuration from CAPM3 here.

It remains to be seen what sort of mess dependabot will do in 1.7 and 1.6 branches as it did in CAPM3, beware for the storm :)

tuminoid:
/override metal3-centos-e2e-integration-test-main  metal3-ubuntu-e2e-integration-test-main

PR #562, opened 2024-06-17
🌱 bump k8s.io/code-generator to v0.30.2

tuminoid:
Do a manual bump to of code-generator so we can control the go directive in hack/tools/go.mod and not inject toolchain directive there.

Supersedes: https://github.com/metal3-io/ip-address-manager/pull/561


kashifest (commented):
/approve

PR #639, opened 2024-08-13
🌱 Uplift CAPI to 1.8.0 and Drop caBundle from CRDs

adilGhaffarDev:
Uplift CAPI to 1.8.0 and Drop caBundle from CRDs to support upcoming Kubernetes 1.31.


adilGhaffarDev:
cc @Sunnatillo @smoshiur1237

PR #721, opened 2024-10-31
🌱 Enable TLS 1.3 flag in IPAM

smoshiur1237:
PR will add the` --tls-min-version` flag value to `VersionTLS13` to use TLS 1.3.


Sunnatillo (approved):
/approve

tuminoid:
Original thought was to add it as a kustomize option, but OTOH, we did make it default in 1.6 already, so this is more like a regression fix, so its fine.

PR #728, opened 2024-11-04
🌱 Use OSV-Scanner release binary and Reporter action workflow instead of pointing to a file

kashifest:
Pointing to the workflow file in OSV scanner repository leads to use a commit sha which doesn't correspond to the osv scanner's release tag sha. Hence it is cumbersome to manually check and update the file's sha everytime a new release is there. This PR uses OSV scanner binary and Reported action workflow from release and the sha pinning also corresponds to the release tag. We also tried to use OSV scanner workflow instead of installing the binary but the workflow build the OSV scanner container and the container doesnt have go binary, hence go call analysis fails. Hence we install and use the binary instead.

Also changing the schedule scan time to 6 hours earlier to make it easier to debug the issue.

Signed-off-by: Kashif Khan <kashif.khan@est.tech>

### DIFF
--- a/.github/workflows/osv-scanner-scan.yml
+++ b/.github/workflows/osv-scanner-scan.yml
@@ -5,7 +5,7 @@ name: OSV-Scanner Scan
 
 on:
   schedule:
-  - cron: "12 12 * * 1"
+  - cron: "12 6 * * 1"
 
 # Restrict jobs in this workflow to have no permissions by default; permissions
 # should be granted per job as needed using a dedicated `permissions` block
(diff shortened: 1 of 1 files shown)


.github/workflows/osv-scanner-scan.yml
```diff
@@ -18,4 +18,29 @@ jobs:
...
+        uses: google/osv-scanner/actions/reporter@b13f37e1a1e4cb98556c1d34cd3256a876929be1 # v1.9.1
+        with:
+           scan-args: |-
+            --output=results.sarif
+            --new=results.json
+            --gh-annotations=false
+      - name: Upload SARIF file
+        uses: github/codeql-action/upload-sarif@v2
+        with:
+          sarif_file: results.sarif
```
tuminoid, line 46: Missing linefeed.

.github/workflows/osv-scanner-scan.yml
```diff
@@ -18,4 +18,29 @@ jobs:
...
+        id: vars
+        run: echo "go_version=$(make go-version)" >> $GITHUB_OUTPUT
+      - name: Set up Go
+        uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
+        with:
+          go-version: ${{ steps.vars.outputs.go_version }}
+      - name: Install OSV Scanner
+        run: go install github.com/google/osv-scanner/cmd/osv-scanner@b13f37e1a1e4cb98556c1d34cd3256a876929be1 # v1.9.1
+      - name: Run OSV Scanner
+        run: osv-scanner --format json --output results.json -r --skip-git ./
```
tuminoid, line 34:
```suggestion
        run: osv-scanner scan --format json --output results.json --recursive --skip-git ./
```

1. Lets use proper subcommand for scanning, otherwise it'll eventually break when they remove the "scan is the default action" fallback.
2. All other args are long form so let's use long form for recursive as well.

.github/workflows/osv-scanner-scan.yml
```diff
@@ -18,4 +18,29 @@ jobs:
       contents: read # to fetch code (actions/checkout)
       security-events: write # for uploading SARIF files
     if: ${{ github.repository == 'metal3-io/ip-address-manager' && github.event_name == 'schedule' }}
```
tuminoid, line 20:
```suggestion
    if: ${{ github.repository == 'metal3-io/ip-address-manager' }}
```
We only trigger from schedule, we can remove extra check.

tuminoid (commented):
Some minor improvement suggestions.

kashifest:
/override metal3-centos-e2e-integration-test-main  metal3-ubuntu-e2e-integration-test-main

tuminoid (commented):
/approve

Rozzii (commented):
/lgtm

PR #734, opened 2024-11-07
🌱 Automating release creation

adilGhaffarDev:
This PR is
- Fixing release automation according to this issue: https://github.com/metal3-io/cluster-api-provider-metal3/issues/1923 
- fixing release note generation
- updating documentation
Fixes #


adilGhaffarDev:
documentation needs to be updated I am working on it.

Makefile
```diff
@@ -306,7 +306,7 @@ release-manifests: $(KUSTOMIZE) $(RELEASE_DIR) ## Builds the manifests to publis
 
 .PHONY: release-notes
 release-notes: $(RELEASE_NOTES_DIR) $(RELEASE_NOTES)
-	go run ./hack/tools/release/notes.go --from=$(PREVIOUS_TAG) > $(RELEASE_NOTES_DIR)/$(RELEASE_TAG).md
+	go run ./hack/tools/release/notes.go --from=$(PREVIOUS_TAG) --releaeTag=$(RELEASE_TAG) > $(RELEASE_NOTES_DIR)/$(RELEASE_TAG).md
```
tuminoid, line 309: typo

.github/workflows/release.yaml
```diff
@@ -1,39 +1,114 @@
...
+      - name: Get release version
+        id: release-version
+        run: |
+            if [[ ${{ steps.changed-files.outputs.all_changed_files_count }} != 1 ]]; then
+              echo "1 release notes file should be changed to create a release tag, found ${{ steps.changed-files.outputs.all_changed_files_count }}"
+              exit 1
+            fi
+            for changed_file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+              export RELEASE_VERSION=$(echo "${changed_file}" | grep -oP '(?<=/)[^/]+(?=\.md)')
+              echo "RELEASE_VERSION=$RELEASE_VERSION" >> $GITHUB_ENV
```
tuminoid, line 36:
```suggestion
              echo "RELEASE_VERSION=${RELEASE_VERSION}" >> ${GITHUB_ENV}
```
Fix all the curlys in this file please.

.github/workflows/release.yaml
```diff
@@ -1,39 +1,114 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # tag=v5.0.2
+        with:
+          go-version: ${{ env.go_version }}
+      - name: generate release artifacts
+        run: |
+          make release
+      - name: get release notes
+        run: |
+          curl -L "https://raw.githubusercontent.com/${{ github.repository }}/main/CHANGELOG/${{ env.RELEASE_TAG }}.md" \
```
tuminoid, line 106: This implies all changelogs are put into main, and not respective release branches. I guess that works as well.

Makefile
```diff
@@ -291,7 +291,7 @@ kind-reset: ## Destroys the "ipam" kind cluster.
 
 RELEASE_TAG ?= $(shell git describe --abbrev=0 2>/dev/null)
 RELEASE_DIR := out
-RELEASE_NOTES_DIR := releasenotes
+RELEASE_NOTES_DIR := CHANGELOG
```
tuminoid, line 294: Does this dir actually matter what its called?

.github/workflows/release.yaml
```diff
@@ -1,39 +1,114 @@
...
-    tags:
-    - "v*"
+    branches:
+      - main
+    paths:
+      - 'CHANGELOG/*.md'
 
-permissions: {}
+permissions:
+  contents: write # Allow to push a tag, create a release branch and publish a draft release.
```
tuminoid, line 12: Do add it to top-level permissions. Put it into specific job that needs that permission.

.github/workflows/release.yaml
adilGhaffarDev: fixed

.github/workflows/release.yaml
adilGhaffarDev: Yes PR will be opened and merge on the main branch.

Makefile
adilGhaffarDev: No, it does not matter, we have to just make sure that this folder name is the same as the one we are checking in the workflow script. If we change it here we have to change in workflow script too. 
In my opinion CHANGELOG is ok, but we can change it if you have some other suggestions.

.github/workflows/release.yaml
adilGhaffarDev: there are 2 jobs
- push_release_tags
- release
And both need this permission. 
Can steps have different permissions?

tuminoid: Yes. Put minimum per job.

Adding to top-level permissions leads to 
1. excessive permissions when job needs differ
2. applies those top-level permissions to new jobs added which might not need any

Permissions is thus best applied per job.

Makefile
tuminoid: Well, we have releasenotes dir already, which currently is just a placeholder. IMO it is just as valid name as changelog would be, without requiring any change here.

.github/workflows/release.yaml
```diff
@@ -1,39 +1,114 @@
...
+            echo "Branch ${RELEASE_BRANCH} already exists"
+            git checkout "${RELEASE_BRANCH}"
+          else
+            git checkout -b "${RELEASE_BRANCH}"
+            git push origin "${RELEASE_BRANCH}"
+            echo "Created branch ${RELEASE_BRANCH}"
+          fi
+      - name: Validate tag does not already exist
+        run: |
+          if [[ $(git tag -l ${RELEASE_VERSION}) ]]; then
```
tuminoid, line 68: Let's make it clear what it tries to achieve.

```suggestion
          if [[ -n "$(git tag -l "${RELEASE_VERSION}")" ]]; then
```

.github/workflows/release.yaml
adilGhaffarDev: done

Makefile
adilGhaffarDev: changed it back to `releasenotes`, also removed `releasenotes` from gitignore.

.github/workflows/release.yaml
```diff
@@ -1,39 +1,117 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # tag=v5.0.2
+        with:
+          go-version: ${{ env.go_version }}
+      - name: generate release artifacts
+        run: |
+          make release
+      - name: get release notes
+        run: |
+          curl -L "https://raw.githubusercontent.com/${{ github.repository }}/main/CHANGELOG/${{ env.RELEASE_TAG }}.md" \
```
tuminoid, line 109: this is still pointing to CHANGELOG

hack/tools/go.mod
```diff
@@ -1,72 +1,79 @@
...
-go 1.22.3
+go 1.22.7
 
-require (
-	github.com/jteeuwen/go-bindata v3.0.7+incompatible
-	k8s.io/code-generator v0.30.6
-	sigs.k8s.io/controller-tools v0.13.0
-	sigs.k8s.io/testing_frameworks v0.1.2
-)
+toolchain go1.22.9
```
tuminoid, line 5: We don't want toolchain directives. Just put 1.22.9 directly to go directive, and remove this line.

tuminoid: We should also bump go to 1.22.9 or 1.23.x. I think generate is failing because of that.

adilGhaffarDev: let's do this in separate PR, I was able to fix the issue without needing to do that.

.github/workflows/release.yaml
adilGhaffarDev: fixed

tuminoid:
Also, let's design this so that we can replicate it to all releasable repos here. For other workflows, we have moved the boilerplate code to project-infra and then simply call it from repo being released with correct params. Makes much less maintenance in the long run, when updates need to go in one repo only. 

Here we of course have the release note code in the repo, which might make it difficult, but at least check if above is possible?

adilGhaffarDev:
> Also, let's design this so that we can replicate it to all releasable repos here. For other workflows, we have moved the boilerplate code to project-infra and then simply call it from repo being released with correct params. Makes much less maintenance in the long run, when updates need to go in one repo only.

Yes, @mquhuy was also suggesting that. I will check.

adilGhaffarDev:
As discussed, let's proceed with merging this change and test it in the upcoming patch releases. Once verified, I'll open a pull request in project-infra to add a common workflow script for IPAM, CAPM3, and BMO.
cc @tuminoid @Rozzii

adilGhaffarDev:
@peppi-lotta @smoshiur1237 @Sunnatillo please review

.github/workflows/release.yaml
```diff
@@ -1,39 +1,117 @@
...
+# This code is borrowed from https://github.com/kubernetes-sigs/cluster-api/blob/main/.github/workflows/release.yaml
+name: Create Release
 
 on:
   push:
-    # Sequence of patterns matched against refs/tags
-    tags:
-    - "v*"
+    branches:
+      - main
```
tuminoid, line 7: Indents in the file are wrong, please fix.

```suggestion
    - main
```

.github/workflows/release.yaml
```diff
@@ -1,39 +1,117 @@
...
+              if [[ "${RELEASE_VERSION}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$ ]]; then
+                echo "Valid semver: ${RELEASE_VERSION}"
+              else
+                echo "Invalid semver: ${RELEASE_VERSION}"
+                exit 1
+              fi
+            done
+      - name: Determine the release branch to use
+        run: |
+            if [[ ${RELEASE_VERSION} =~ beta ]] || [[ ${RELEASE_VERSION} =~ alpha ]]; then
```
tuminoid, line 48: This makes us commit to releasing where alpha and beta are tags on main, but RC and actual release tag has to be in release branch.

Do we REALLY want to commit to?

.github/workflows/release.yaml
```diff
@@ -1,39 +1,117 @@
...
+        run: |
+          if [[ -n "$(git tag -l "${RELEASE_VERSION}")" ]]; then
+            echo "Tag ${RELEASE_VERSION} already exists, exiting"
+            exit 1
+          fi
+      - name: Create Release Tag
+        run: |
+          git config user.name "${GITHUB_ACTOR}"
+          git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+          git tag -a ${RELEASE_VERSION} -m ${RELEASE_VERSION}
```
tuminoid, line 77: These tags are wrong?

They should match per repo, and for IPAM we have:
- main tag
- api tag
- test tag

.github/workflows/release.yaml
adilGhaffarDev: I think this is fine, this is something that is followed in the Kubernetes release, and CAPI also followed them.
What are the potential drawbacks of this? 
Keep creating betas until we have the code freeze and release RCs afterward.

tuminoid: It is just commitment we need to agree to. 

We have not been doing too many branching RCs in the past, and in general not many alpha/betas either. We then can't do RC releases from main, but must cut the branch and fork the work at that point. We know branching for minor is not trivial with out multi-repo setup, so I just want us to make informed decision that this piece of code will dictate that process going forward.

For example, pushing a branch here does not mean we can test it in anyway unlike CAPI in a single repository setup.

.github/workflows/release.yaml
tuminoid: seems fixed now

.github/workflows/release.yaml
tuminoid: As discussed offline, let's mark this
/hold
For agreement next week when people are back from kubecon etc

Sunnatillo (approved):
/approve

.github/workflows/release.yaml
```diff
@@ -1,39 +1,117 @@
...
+    - name: generate release artifacts
       run: |
         make release
+    - name: get release notes
+      run: |
+        curl -L "https://raw.githubusercontent.com/${{ github.repository }}/main/releasenotes/${{ env.RELEASE_TAG }}.md" \
+        -o "${{ env.RELEASE_TAG }}.md"
     - name: Release
-      uses: softprops/action-gh-release@e7a8f85e1c67a31e6ed99a94b41bd0b71bbee6b8 # v2.0.9
+      uses: softprops/action-gh-release@c062e08bd532815e2082a85e87e3ef29c3e6d191 # tag=v2.0.8
```
mquhuy, line 112: Don't downgrade this

.github/workflows/release.yaml
```diff
@@ -1,39 +1,117 @@
...
-    if: github.repository == 'metal3-io/ip-address-manager'
     steps:
-    - name: Export RELEASE_TAG var
-      run: echo "RELEASE_TAG=${GITHUB_REF:10}" >> $GITHUB_ENV
+    - name: Set env
+      run: echo "RELEASE_TAG=${RELEASE_TAG}" >> ${GITHUB_ENV}
+      env:
+        RELEASE_TAG: ${{needs.push_release_tags.outputs.release_tag}}
     - name: checkout code
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
mquhuy, line 24: Don't downgrade this?

.github/workflows/release.yaml
```diff
@@ -1,39 +1,117 @@
...
+        git push origin api/${RELEASE_VERSION}
+        echo "Created tags ${RELEASE_VERSION} and api/${RELEASE_VERSION}"
+  release:
+    name: create draft release
+    runs-on: ubuntu-latest
+    needs: push_release_tags
     permissions:
       contents: write
-
-    if: github.repository == 'metal3-io/ip-address-manager'
```
mquhuy, line 19: Maybe keep this (and add it to the other job too)? We may never push to forks' main, but who knows :-?

.github/workflows/release.yaml
```diff
@@ -1,39 +1,117 @@
...
+    - name: Set env
+      run: echo "RELEASE_TAG=${RELEASE_TAG}" >> ${GITHUB_ENV}
+      env:
+        RELEASE_TAG: ${{needs.push_release_tags.outputs.release_tag}}
     - name: checkout code
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # tag=v4.1.7
       with:
         fetch-depth: 0
-    - name: Install go
```
mquhuy, line 27: This action is also downgraded.

mquhuy (commented):
Thanks @adilGhaffarDev , this looks like a very good change!

I had some nits from initial reviews, may come back with more later if I have time.

kashifest:
/hold cancel
ready for review

docs/releasing.md
tuminoid, line 1: This needs more revamping. Permissions section for example is now invalid. Reposotiry setup also needs some cleaning.

Release notes section also needs complete rewrite.

docs/releasing.md
```diff
@@ -53,22 +53,27 @@ Clone the repository:
...
+IMPORTANT_NOTE:
+   - The commit and PR title should be 🚀 Release v1.x.y.
+   - Important! The commit should only contain the release notes file, nothing
+    else, otherwise automation will not work.
+
+- Ask maintainers and release team members to review your pull request.
+
+Once PR is merged it triggers three things:
 
 - GitHub action workflow for automated release process creates a draft release
```
tuminoid, line 78: This is also not right.

docs/releasing.md
```diff
@@ -83,6 +88,11 @@ This triggers two things:
   [Quay tags page](https://quay.io/repository/metal3-io/ip-address-manager?tab=tags).
   If the release tag build is not visible, check if the action has failed and
   retrigger as necessary.
+- GH job `create draft release` creates draft release. Don't publish the release
+  until release tag is visible in
+  [Quay tags page](https://quay.io/repository/metal3-io/ip-address-manager?tab=tags).
+  If the release tag build is not visible, check if the action has failed and
+  retrigger as necessary.
 
 We also need to create one or more tags for the Go modules ecosystem:
```
tuminoid, line 97: This is also obsolete now, or needs to be reworded in a way that explains how the new automation does it.

tuminoid (commented):
/hold

We need to have documents in more finished state. I realize we need further tuning after we've done a release, but at least the steps should be corrected. Now there is too much old stuff that should not be done by release team at all.

docs/releasing.md
adilGhaffarDev: agree, I will make these changes.

docs/releasing.md
```diff
@@ -47,75 +32,63 @@ IPAM uses [semantic versioning](https://semver.org).
...
+IMPORTANT_NOTE:
+   - The commit and PR title should be 🚀 Release v1.x.y.
+   - Important! The commit should only contain the release notes file, nothing
+     else, otherwise automation will not work.
+
+- Ask maintainers and release team members to review your pull request.
+
+Once PR is merged it triggers three things:
+
+- Github action `push_release_tags` will create and push the tags. This action
```
tuminoid, line 74: There is no action with this name. All were adding is `release.yml` with `Create release` name.

What you're referring are jobs in the workflow.

docs/releasing.md
```diff
@@ -47,75 +32,63 @@ IPAM uses [semantic versioning](https://semver.org).
...
+  will also create release branch if its missing and release is `rc` or minor.
+- Github action `build-images-action` starts building release image with the
+  release tag in Jenkins, and it gets pushed to Quay. Make sure the release tag
+  is visible in
+  [Quay tags page](https://quay.io/repository/metal3-io/ip-address-manager?tab=tags).
+  If the release tag build is not visible, check if the action has failed and
+  retrigger as necessary.
+- GH job `create draft release` creates draft release. Don't publish the release
```
tuminoid, line 82: Move this above build-images-action so jobs within same workflow are discussed together.

docs/releasing.md
adilGhaffarDev: fixed.

tuminoid:

tuminoid (commented):
/lgtm

Lets take this baby for a test ride!

PR #741, opened 2024-11-12
:seedling: use make go-version instead of Dockerfile parsing in verify-release.sh

tuminoid:
Make use of "make go-version" instead of Dockerfile parsing to do golang version related operations. This alters base image version check and also osv-scanner now has a new functionality to override the detected Go version, which we will use via "make go-version" as well.

This cleans up code quite a bit, and does not rely on excessive hacks for the vulnerability scanning.

### DIFF
--- a/hack/verify-release.sh
+++ b/hack/verify-release.sh
@@ -439,29 +439,15 @@ verify_container_images()
     echo -e "Done\n"
 }
 
-_get_golang_version_from_dockerfile()
-{
-    # read golang version from Dockerfile and return
-    local image_and_tag image image_and_tag_without_sha tag tag_minor
-
-    image_and_tag="$(grep "^ARG BUILD_IMAGE=" Dockerfile | cut -f2 -d=)"
-    image="${image_and_tag/:*}"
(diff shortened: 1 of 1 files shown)


hack/verify-release.sh
```diff
@@ -626,42 +612,24 @@ verify_module_releases()
...
+    local go_version config_file=".osv-scanner.toml"
+
     echo "Verifying vulnerabilities ..."
 
-    _mutate_gomod_files_for_osv_scanner
+    go_version="$(make go-version)"
+    echo "GoVersionOverride = \"${go_version}\"" > "${config_file}"
+    "${OSVSCANNER_CMD[@]}" scan \
+        --skip-git --recursive \
+        --config=.osv-scanner.toml \
```
lentzi90, line 626:
```suggestion
        --config="${config_file}" \
```

lentzi90 (commented):
Nice!

hack/verify-release.sh
tuminoid: Fixed

lentzi90 (commented):
/approve

tuminoid:
/test all
Testing prow

PR #775, opened 2024-12-05
:bug: Make image build on releases

### LINKED ISSUE #772: With new release automation, container images workflow is no longer trigged by tag creation
After https://github.com/metal3-io/ip-address-manager/pull/734 was merged, release tags are created by this workflow. This means build-images-action is no longer triggering by tag creation, due Github side security precaution to avoid loops. 

We need to fix it. Some ideas:
- `workflow_run` - change the triggering mechanism, issue is how the tag data is passed
- `workflow_dispatch` - maybe we change both a bit so `uses:` can be integrated as part of release

It does not seem that adding more permissions like `actions: write` etc helps.

mquhuy:

Fixes #772 

This should be merged after https://github.com/metal3-io/project-infra/pull/927

### DIFF
--- a/.github/workflows/build-images-action.yml
+++ b/.github/workflows/build-images-action.yml
@@ -6,20 +6,38 @@ permissions:
 on:
   push:
     branches:
-    - 'main'
-    - 'release-*'
+    - "main"
+    - "release-*"
     tags:
-    - 'v*'
+    - "v*"
+  workflow_run:
+    workflows:
+    - "Create Release"
+    types:
+    - completed
 
 jobs:
   build_ipam:
-    name: Build IPAM container image
-    if: github.repository == 'metal3-io/ip-address-manager'
-    uses: metal3-io/project-infra/.github/workflows/container-image-build.yml@main
-    with:
-      image-name: 'ip-address-manager'
-      pushImage: true
-    secrets:
-      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
-      QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
-      SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set ref
+      id: set_ref
+      run: |
+        # Check if release_tag exists, if so, set ref to "refs/heads/${{ release_tag }}", otherwise use github.ref
+        if [[ -n "${{ needs.create-release.outputs.release_tag }}" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/build-images-action.yml
```diff
@@ -10,6 +10,11 @@ on:
     - 'release-*'
     tags:
     - 'v*'
+  workflow_run:
```
tuminoid, line 13: This triggers the build images action workflow, yes, but it does not pass context why it was triggered, so it does not know that it needs to build tags.

mquhuy: Thanks. I've now added https://github.com/metal3-io/project-infra/pull/927 to account for that.
But now I'm wondering if the "Create Release" wf should be made reusable :-? Wdyt about that?

tuminoid: You mean moving it to project-infra? Yes it should, but only after we've ironed out the kinks.

.github/workflows/build-images-action.yml
```diff
@@ -6,20 +6,38 @@ permissions:
...
-      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
-      QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
-      SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set ref
+      id: set_ref
+      run: |
+        # Check if release_tag exists, if so, set ref to "refs/heads/${{ release_tag }}", otherwise use github.ref
+        if [[ -n "${{ needs.create-release.outputs.release_tag }}" ]]; then
```
tuminoid, line 27: What sets this?

mquhuy: It's inside the "Create Release" workflow.

tuminoid: Yes, but how does it end up being set here? What populates the `needs.` so it can be used here?

mquhuy: It's this line: https://github.com/metal3-io/ip-address-manager/blob/main/.github/workflows/release.yaml#L19

tuminoid (commented):
/approve

Technically it looks solid, but these inter-action relationships are so flaky, that I guess we'll find out when we cut next beta or RC with the new automation if it does what it we think it should be doing :) 

Thanks @mquhuy !

tuminoid:

lentzi90 (commented):
/lgtm

PR #749, opened 2024-11-20
:seedling: bump Go to 1.23.3 and golangci-lint to 1.60.3

tuminoid:
Bump Go to 1.23.3 and golangci-lint to 1.60.3, which supports Go 1.23.

Update golangci-lint config, remove deprecations there.

Fix new issues found by golangci-lint.

Fixes: #716

PR #777, opened 2024-12-09
:seedling: Lint hack/tools

### LINKED ISSUE #767: Golangci-lint does not lint release note tools
Golangci-lint does not lint release note tools, only root + api dirs.

Add other directories with .go files to linter config and then fix the go files if linter fails on them.

/kind bug

wallchristopher:

Added the hack/tools/release directory to be linted like the rest of the repo. Attempted to do the hack/tools directory but ran into typecheck errors stating that the imports were not packages. Per golangci-lint's docs, [you can't skip or ignore these errors](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors). Instead of modifying the import, I opted for just scanning the actual code.

Something to note, some of the keys within the [.golangci.yml](https://github.com/metal3-io/ip-address-manager/blob/main/.golangci.yaml) at not valid such as the [issues.skip-dirs](https://github.com/metal3-io/ip-address-manager/blob/main/.golangci.yaml#L94) and issues.skip-files. When playing around with the configuration they were a gotcha that got me.

Addressed issues identified by the linters.

**Tested via**:

local: within hack/tools/release dir `golangci-lint run --build-tags=tools --modules-download-mode=readonly`
act: repo root `act -j golangci`

Fixes #767

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -17,6 +17,10 @@ jobs:
         working-directory:
         - ""
         - api
+        - hack/tools/release
+        include:
+        - working-directory: hack/tools/release
+          additional-args: --build-tags=tools --modules-download-mode=readonly
 
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
@@ -32,4 +36,4 @@ jobs:
       with:
         version: v1.60.3
         working-directory: ${{matrix.working-directory}}
-        args: --timeout=15m
+        args: --timeout=15m ${{matrix.additional-args}}
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,7 @@ $(KUSTOMIZE): $(TOOLS_DIR)/go.mod
 lint: $(GOLANGCI_LINT) ## Lint codebase
 	$(GOLANGCI_LINT) run -v $(GOLANGCI_LINT_EXTRA_ARGS) --timeout=10m
 	cd $(APIS_DIR); ../$(GOLANGCI_LINT) run -v $(GOLANGCI_LINT_EXTRA_ARGS) --timeout=10m
+	cd $(TOOLS_DIR)/release; ../../../$(GOLANGCI_LINT) run -v --build-tags=tools --modules-download-mode=readonly $(GOLANGCI_LINT_EXTRA_ARGS) --timeout=10m
 
 .PHONY: lint-fix
 lint-fix: $(GOLANGCI_LINT) ## Lint the codebase and run auto-fixers if supported by the linter
@@ -150,6 +151,7 @@ lint-fix: $(GOLANGCI_LINT) ## Lint the codebase and run auto-fixers if supported
 lint-full: $(GOLANGCI_LINT) ## Run slower linters to detect possible issues
 	$(GOLANGCI_LINT) run -v --fast=false
 	cd $(APIS_DIR); ../$(GOLANGCI_LINT) run -v --fast=false --timeout=30m
+	cd $(TOOLS_DIR)/release; ../../../$(GOLANGCI_LINT) run -v --fast=false --build-tags=tools --modules-download-mode=readonly --timeout=30m
 
 ## --------------------------------------
 ## Generate
--- a/hack/tools/release/notes.go
+++ b/hack/tools/release/notes.go
@@ -20,7 +20,6 @@ limitations under the License.
 package main
 
 import (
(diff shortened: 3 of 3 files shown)


tuminoid:
> **What this PR does / why we need it**:
> 
> Added the hack/tools/release directory to be linted like the rest of the repo. Attempted to do the hack/tools directory but ran into typecheck errors stating that the imports were not packages. Per golangci-lint's docs, [you can't skip or ignore these errors](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors). Instead of modifying the import, I opted for just scanning the actual code.

/cc @kashifest @lentzi90 
WDYT?

> Something to note, some of the keys within the [.golangci.yml](https://github.com/metal3-io/ip-address-manager/blob/main/.golangci.yaml) at not valid such as the [issues.skip-dirs](https://github.com/metal3-io/ip-address-manager/blob/main/.golangci.yaml#L94) and issues.skip-files. When playing around with the configuration they were a gotcha that got me.

The config was [recently moved](https://github.com/metal3-io/ip-address-manager/pull/749) there as the old location was deprecated, and this should be the correct place for the configs. I don't see any complaints for golangci-lint, using "make lint". 

> Addressed issues identified by the linters.

Those look legit changes to me, and especially the unused/if-else problems alerts me of this linting issue in the first place.

> **Tested via**:
> 
> local: within hack/tools/release dir `golangci-lint run --build-tags=tools --modules-download-mode=readonly` act: repo root `act -j golangci`

"make lint"  should be linting this, but currently it does not. If I checkout the old notes.go from main which does not have any fixes, it will not catch any of them. This needs to be fixed.

.github/workflows/golangci-lint.yml
```diff
@@ -17,6 +17,10 @@ jobs:
         working-directory:
         - ""
         - api
+        - hack/tools/release
```
tuminoid, line 20: This does not get linted using this config.

.github/workflows/golangci-lint.yml
```diff
@@ -17,6 +17,10 @@ jobs:
...
+        include:
+          - working-directory: hack/tools/release
+            additional-args: --build-tags=tools --modules-download-mode=readonly
```
tuminoid, line 23:
```suggestion
        - working-directory: hack/tools/release
          additional-args: --build-tags=tools --modules-download-mode=readonly
```
All YAML is indented in k8s style.

lentzi90 (commented):
Looks solid!
I think the commits should be squashed. It is not necessary to have one for each little fix. Please fix the comments by @tuminoid also, then I think this is good to merge

wallchristopher:
@tuminoid thank you for pointing out the `make lint`. Updated to now lint the tools directory.

> The config was https://github.com/metal3-io/ip-address-manager/pull/749 there as the old location was deprecated, and this should be the correct place for the configs. I don't see any complaints for golangci-lint, using "make lint"

It might just be my text-editor. Golangci-lint's docs don't mention skip-dirs / skip-files anymore. [Digging within the golangci-lint's code](https://github.com/golangci/golangci-lint/blob/master/pkg/config/run.go#L30) it seems that it still works but with a deprecation notice. Only wanted to bring it up.
image

@lentzi90 Appreciate it, squashed the commits into one.

.github/workflows/golangci-lint.yml
wallchristopher: I'm not sure what you mean. The /.golangci-lint.yml config is used when golangci-lint is run.

If meaning when running `make lint`, I completely missed adding it. Thank you for bringing that up. Should be good picked up now.

tuminoid:
> It might just be my text-editor. Golangci-lint's docs don't mention skip-dirs / skip-files anymore. [Digging within the golangci-lint's code](https://github.com/golangci/golangci-lint/blob/master/pkg/config/run.go#L30) it seems that it still works but with a deprecation notice. Only wanted to bring it up. 

You're correct, it is deprecated. I wonder why it doesn't complain about those after they were once fixed in https://github.com/metal3-io/ip-address-manager/pull/749 and then it did complain about them.

 There are some other settings deprecated as well, so might as well fix them in a follow-up PR. And bump to latest version also. Previous bump was done minimally to just get support for Go 1.23.

tuminoid:

tuminoid (commented):
/approve

This is good improvement. Thank you for contributing @wallchristopher !

Rozzii (commented):
/lgtm

Rozzii:
Thanks @wallchristopher

PR #785, opened 2024-12-12
:seedling: Bump Golang to v1.23.4

peppi-lotta:
**What this PR does / why we need it**: Bump Golang from v1.23.3 to v1.23.4


tuminoid (commented):
/approve

PR #787, opened 2024-12-12
🌱 Introduce ReconcileError with Transient and Terminal Error type

Sunnatillo:
This PR:

Introduces ReconcileError with Transient and Terminal Error type
Removes RequeueAfterError (as it was depricated in CAPI and CAPM3)

ReconcileError represents an generic error of Reconcile loop. errorType indicates what type of action is required to recover. It can take two values:
a. Transient - Can be recovered , will be requeued after.
b. Terminal - Cannot be recovered, will not be requeued.

When error is not ReconcileError it will be returned to Reconciler.


ipam/reconcile_error.go
```diff
@@ -0,0 +1,85 @@
+/*
```
Rozzii, line 1: I assume from the license message, that this file and the related test file were copied.
Is there a possibility to find these Error types in some library that we could pull in as a dependency instead of duplicating  this code?

If this is not copied but written by you, then my question is that why are you adding the K8s author copyright?

Rozzii: I know other files have this copyright here, I will start a discussion with the community related to that also.

Rozzii (commented):
/test metal3-ubuntu-e2e-integration-test-main

tuminoid (commented):
We did similar for CAPM3 too, some time back right?

/cc @kashifest @lentzi90

ipam/reconcile_error.go
Sunnatillo: I have copied from capm3. I am pretty much sure I have written those code. What would be proper copyright in this case?

tuminoid: We have used at least `# Copyright 2023 The Metal3 Authors.` elsewhere.

tuminoid:
/hold
until the copyright is corrected.

ipam/reconcile_error.go
Sunnatillo: Fixed now. I used 2024 as PR is opened in 2024

lentzi90 (commented):
/approve
Related to this, I guess we should also think about following CAPI by using conditions instead of `ErrorMessage`, right?

ipam/reconcile_error.go
Rozzii: Thanks @Sunnatillo !

Rozzii (commented):
/lgtm

Sunnatillo:
> /approve Related to this, I guess we should also think about following CAPI by using conditions instead of `ErrorMessage`, right?

yes, I will checl what beta2 is bringing and trying to align with it

tuminoid:
This is currently held for copyright issue, which is now fixed. But then

> yes, I will checl what beta2 is bringing and trying to align with it

Are you going to change this PR still, or merge this and then change it later?

Unhold if latter.

Sunnatillo:
> This is currently held for copyright issue, which is now fixed. But then
> 
> > yes, I will checl what beta2 is bringing and trying to align with it
> 
> Are you going to change this PR still, or merge this and then change it later?
> 
> Unhold if latter.

No, this is different task. 
/unhold

PR #839, opened 2025-01-02
:seedling: Add build container image as a job in release wf

### LINKED ISSUE #772: With new release automation, container images workflow is no longer trigged by tag creation

mquhuy:
This adds build-container-image as a job in the release wf. After the release note draft is created, the container image will be created.

Fixes #772

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -116,3 +116,17 @@ jobs:
         files: out/*
         body_path: ${{ env.RELEASE_TAG }}.md
         tag_name: ${{ env.RELEASE_TAG }}
+
+  build_ipam:
+    needs: push_release_tags
+    name: Build IPAM container image
+    if: github.repository == 'metal3-io/ip-address-manager'
+    uses: metal3-io/project-infra/.github/workflows/container-image-build.yml@main
+    with:
+      image-name: "ip-address-manager"
+      pushImage: true
+      ref: ${{ needs.push_release_tags.outputs.release_tag }}
+    secrets:
+      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+      QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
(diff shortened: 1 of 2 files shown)


peppi-lotta:
\lgtm

.github/workflows/release.yaml
```diff
@@ -116,3 +116,17 @@ jobs:
...
+
+  build_ipam:
+    needs: push_release_tags
+    name: Build IPAM container image
+    if: github.repository == 'metal3-io/ip-address-manager'
+    uses: metal3-io/project-infra/.github/workflows/container-image-build.yml@main
+    with:
+      image-name: "ip-address-manager"
+      pushImage: true
+      ref: ${{ needs.push_release_tags.outputs.release_tag }}
```
adilGhaffarDev, line 128:
```suggestion
      ref: ${{ env.RELEASE_TAG }}
```
We can also use env.RELEASE_TAG here right?

mquhuy: Not really. The `release` job isn't configured to run before `build_ipam`. They could be running in parallel, or in any order. Only the `push_release_tag` is guaranteed to run first. Also, I'm not sure if the env is maintained between the jobs.

Is there any reason for using env instead of outputs?

adilGhaffarDev:
cc @Rozzii @kashifest

kashifest:
@mquhuy wont we need this for release-1.9 patch releases anyhow or do we do the releases from main branch always?

PR #942, opened 2025-03-21
🌱  Bump golangci-lint to v1.64.7

kashifest:
This PR also bumps golangci-lint-action to 6.5.2. While doing that it removes some of the old unsupported golangci-lint settings.
Signed-off-by: Kashif Khan <kashif.khan@est.tech>


tuminoid (commented):
/approve

tuminoid (commented):
/approve cancel

This has broken in CAPM3 and IRSO, let's fix those and take the fixes into this too

tuminoid (commented):
/approve
/lgtm

It was a messup in the golangci-lint end, related to v2 release.

PR #957, opened 2025-03-26
:seedling: Bump CAPI to 1.10.0-beta.1

peppi-lotta:
**What this PR does / why we need it**: Bump CAPI to 1.10.0-beta.1


kashifest (commented):
/approve

PR #976, opened 2025-04-02
:seedling: do not bump major version of github actions

tuminoid:
Change dependabot config not to try major bumps of github actions. We now have grouped bumps to lessen the noise from dependabot, and if in that group, there is a major bump (like right now a golangci-lint v2 major bump) which isn't compatible as is and requires manual changes, it blocks bumping all the other actions as well.

Also, change the name of the all-github-actions group, as how it is inserted in PR title looks wrong. Plain "github-actions" group name works better there.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,8 +12,12 @@ updates:
   target-branch: main
   ## group all action bumps into single PR
   groups:
-    all-github-actions:
+    github-actions:
       patterns: ["*"]
+  ignore:
+  # Ignore major bumps in main, as it breaks the group bump process
+  - dependency-name: "*"
+    update-types: ["version-update:semver-major"]
   commit-message:
     prefix: ":seedling:"
   labels:
@@ -58,7 +62,7 @@ updates:
   target-branch: release-1.9
   ## group all action bumps into single PR
   groups:
-    all-github-actions:
+    github-actions:
       patterns: ["*"]
   ignore:
   # Ignore major and minor bumps for release branch
@@ -105,7 +109,7 @@ updates:
   target-branch: release-1.8
   ## group all action bumps into single PR
   groups:
-    all-github-actions:
+    github-actions:
       patterns: ["*"]
   ignore:
   # Ignore major and minor bumps for release branch


lentzi90 (commented):
/approve

lentzi90:

kashifest:
Whats the plan with major bumps then? do we plan to do them manually? I somehow feel like we were better of before the grouping.

tuminoid:
> Whats the plan with major bumps then? do we plan to do them manually? I somehow feel like we were better of before the grouping.

Majors usually need manual actions anyways, as they contain breaking changes. We do them manually as before, the rest is grouped nicely into one.

kashifest:
> Majors usually need manual actions anyways, as they contain breaking changes. We do them manually as before, the rest is grouped nicely into one.

Majors were not done manually before when they were not grouped. if they are not grouped, the good thing is they will only fail when major is broken. Otherwise we dont need to keep an eye on all the actions' major releases. I am of the opinion of ungrouping them again.

kashifest:
> I really don't see benefit in merging actions one by one, when we cannot test them in any way, they often need rebasing after one of them is merged etc. They pollute release notes with completely unrelated stuff to what we want to show and are just generally really noisy and basically just busy work to review and merge them.

On the contrary its easy to review and manage. Biggest benefit is you don't need to lookout or miss out when a major bump is out. I can agree to the release note pollution but seems like a lesser issue to me. Reviewing and merging would still be less work than manually check and push PRs IMO.   


> Majors are always done manually, if they break something, there is no way around it. Majors just don't happen very often, but recently we just got few in a row to highlight the point.

Yep, thats what I meant. We do it manually only when it breaks. 

I know major's are not that often out, but still are we ready to keep an eye whenever it is out manually cause it is not just one or two actions?

kashifest:
Its funny though that now I remember when I was copying the config from CAPI initially you thought that grouping was not a good idea and I thought it was and now we are on the opposite side 😄. ts 

Lets hope its not too much manual work for manually tracing major releases. 

/lgtm

tuminoid:
> Its funny though that now I remember when I was copying the config from CAPI initially you thought that grouping was not a good idea and I thought it was and now we are on the opposite side 😄.

Generally it is good practice to review small PRs, but I think we've learned by now that we have little to review here as there is no automation etc, and its just pollution at this point. Live and learn :)

kashifest:
> Normally I would agree that smaller PRs are better to review and good practice, but with actions, Just slapping an `/approve` and `override tests` does not really consitute a review anyways. It is busy work. 

in case of bumps like minor/major, I personally have a tendancy to check the release notes as well. 

>  The only pro I see is the "notification" that there is a new major. For me, the cons outweight the pros by far as I don't see why would we need to be on "lookout" for any majors. We don't actively want to bump majors for actions. We bump them when we need to, as side effect.

Yep, exactly. You will get a notification. You dont need to go lookout for it. Whether we take it or not we could decide by looking at the PR for changes.

Issue #969, opened 2025-03-27
Allocating a specific IP from IpClaim


ksahil12:
**User Story**

As a developer/user I would like to allocate a specific IP from a IpPool while creating a IpClaim to acquire a preferred IP.

**Detailed Description**

From an IPAM perspective, there can be various scenarios where we need a specific IP to be allocated from a given set of IpPools if it is available. 

As per my understanding, currently only way of doing this is to update preAllocations in IpPool CR first and then create IpClaim CR with the same name. This seems to be an extra step for acquiring an IP i.e. 2 API calls, one to update IpPool and then create IpClaim.

**High level suggestion**

1. Adding an IPAddress field in IpClaim: _Address IPAddressStr `json:"address"`_
``Address IPAddressStr `json:"address"``
2. In ippool_manager.go
`func (m *IPPoolManager) allocateAddress(addressClaim *ipamv1.IPClaim,
	addresses map[ipamv1.IPAddressStr]string,
) (ipamv1.IPAddressStr, int, *ipamv1.IPAddressStr, []ipamv1.IPAddressStr, error)`

Above method already checks whether there are any preAllocated IPs in IpPool for a given IpClaim. Here, we can utilize the specific IpAddress provided in step1.

/kind feature

tuminoid:
Hi @ksahil12 !

We have [IPReuse](https://github.com/metal3-io/ip-address-manager/blob/main/docs/api.md#ippool) feature, ie. mapping IPAddressClaims via their name to specific IP addresses in IPPool:

```yaml
  preAllocations:
    claim1: 192.168.0.12
```

Is this not achieving what you want? Can you give more details what is lacking?

ksahil12:
Hi @tuminoid 
Challenge in using above approach is that it does not guarantee allocation of particular IP.
Scenarios:
1. If user have updated preAllocations with the two claims of the same IP, it does not result in an error.
2. If IP is already in use, it does not stop user from requesting the same IP.
3. Now suppose user has updated preAllocations with the IP and subsequently creates IpClaim. Now this may result in any of the error scenarios, where desired behavior won't be executed and since it's a two-step acquiring process, it becomes confusing to address the issue and may result in redundant IpPool preAllocation state.

Some of these issues, I feel can be addressed in webhooks and error handling. But still, I think acquiring specific IP is a common use case and since we already have things in place to check preAllocated IPs while creating IpAddress CR, it can be integrated and handled for addressing these challenges.

[...]

PR #997, opened 2025-04-10
✨ Adding feature for allocating requested IP

ksahil12:
**Requirement**: Allocate a specific to IPAddressClaim or IPClaim when created with special label.
**Solution**: During creation of IPAddressClaim, user can provide a special label i.e. ipAddress: <IP_TO_BE_REQUESTED>. Once IPAddressClaim goes for allocation logic in ip_pool_manager, we can handle the allocation of this IP based on the requestedIP and available IPs.
For a happy path scenario, when allocation is successful, we will create IPAddress with the same IP and update IPPool index with the allocated IP.
For a scenario where user has requested an IP which does not belong to range or not available, we can patch IPAddressClaim CR with the error mentioning the same.

Fixes https://github.com/metal3-io/ip-address-manager/issues/969

### DIFF
--- a/ipam/ippool_manager_test.go
+++ b/ipam/ippool_manager_test.go
@@ -1916,6 +1916,125 @@ var _ = Describe("IPPool manager", func() {
 			expectedGateway: (*ipamv1.IPAddressStr)(ptr.To("192.168.0.1")),
 			expectedPrefix:  24,
 		}),
+		Entry("One pool, with start and existing address, ipAddress annotation present", testCaseAllocateAddress{
+			ipPool: &ipamv1.IPPool{
+				Spec: ipamv1.IPPoolSpec{
+					Pools: []ipamv1.Pool{
+						{
+							Start: (*ipamv1.IPAddressStr)(ptr.To("192.168.0.11")),
+							End:   (*ipamv1.IPAddressStr)(ptr.To("192.168.0.20")),
+						},
+					},
+					Prefix:  24,
+					Gateway: (*ipamv1.IPAddressStr)(ptr.To("192.168.0.1")),
+				},
+			},
+			ipClaim: &ipamv1.IPClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "TestRef",
+					Annotations: map[string]string{
+						IPAddressAnnotation: "192.168.0.16",
+					},
+				},
+			},
+			addresses: map[ipamv1.IPAddressStr]string{
+				ipamv1.IPAddressStr("192.168.0.12"): "bcde",
+				ipamv1.IPAddressStr("192.168.0.11"): "abcd",
+			},
+			expectedAddress: ipamv1.IPAddressStr("192.168.0.16"),
+			expectedGateway: (*ipamv1.IPAddressStr)(ptr.To("192.168.0.1")),
+			expectedPrefix:  24,
+		}),
+		Entry("One pool, with start and existing address, ipAddress annotation present but already acquired", testCaseAllocateAddress{
+			ipPool: &ipamv1.IPPool{
+				Spec: ipamv1.IPPoolSpec{
+					Pools: []ipamv1.Pool{
+						{
+							Start: (*ipamv1.IPAddressStr)(ptr.To("192.168.0.11")),
+							End:   (*ipamv1.IPAddressStr)(ptr.To("192.168.0.20")),
+						},
+					},
+					Prefix:  24,
+					Gateway: (*ipamv1.IPAddressStr)(ptr.To("192.168.0.1")),
+				},
+			},
+			ipClaim: &ipamv1.IPClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "TestRef",
+					Annotations: map[string]string{
+						IPAddressAnnotation: "192.168.0.11",
+					},
+				},
+			},
+			addresses: map[ipamv1.IPAddressStr]string{
+				ipamv1.IPAddressStr("192.168.0.12"): "bcde",
+				ipamv1.IPAddressStr("192.168.0.11"): "abcd",
+			},
+			expectError: true,
+		}),
+
+		Entry("One pool, with start and existing address, ipAddress annotation present and requested ipAddress in preAllocations with conflicting ipclaim name", testCaseAllocateAddress{
+			ipPool: &ipamv1.IPPool{
+				Spec: ipamv1.IPPoolSpec{
+					Pools: []ipamv1.Pool{
+						{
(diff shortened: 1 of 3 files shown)


peppi-lotta:
I tested a situation where pools preAllocations are: 
```
Pre Allocations:
  test1:  192.168.0.12
  test2:  192.168.0.15
  test3:  192.168.0.17
```
and I the applied an IPClaim:
```
kind: IPClaim
metadata:
  name: test2
  namespace: default
  labels:
    ipAddress: 192.168.0.20
spec:
  pool:
    name: pool1
    namespace: default
```
IpClaim then has error `Error Message:  Requested IP not available`, which is not true. Could you add something like "PreAllocation and requested ip address are conflicting" or something along those lines. 

Otherwise  the feature seems to be working as I expected.

peppi-lotta:
Could you also add tests for situation
One pool, with start and existing address, ipAddress label present and requested ipAddress in preAllocations with conflicting ipclaim name
One pool, with start and existing address, ipAddress label present and ipAddress in preAllocations with non-conflicting ipclaim name

It is important to understand how the new feature works with preAllocations as those are curretly an important part of CAPM3 upgrading.

ksahil12:
Thanks for suggestion @peppi-lotta.
Updated the scenario. Also added test cases.

peppi-lotta:
Please rebase your commits into one :)

ksahil12:
@peppi-lotta updated the PR.

peppi-lotta:
/lgtm
This seems to work well and as expected. Good job and thank you for contributing!

ksahil12:
@peppi-lotta thanks for the review.
@tuminoid can you help with approvals as it's mentioned above to assign PR to you.

lentzi90 (commented):
Sorry for noticing so late, but could we change this to use annotations instead of labels?
Labels are in general [meant as identifying attributes](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/), not for modifying controller behavior. Annotations on the other hand, are good for this and what is commonly used (e.g. [in ingress controllers](https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md)).

ksahil12:
@lentzi90 agreed. We generally use annotations for providing additional/supportive data. @peppi-lotta Please share your thoughts as well. I can work on changing it once finalized.

peppi-lotta:
Annotation seem more appropriate for this.

ipam/ippool_manager_test.go
```diff
@@ -1916,6 +1916,125 @@ var _ = Describe("IPPool manager", func() {
...
+			},
+			addresses: map[ipamv1.IPAddressStr]string{
+				ipamv1.IPAddressStr("192.168.0.12"): "bcde",
+				ipamv1.IPAddressStr("192.168.0.11"): "abcd",
+			},
+			expectedAddress: ipamv1.IPAddressStr("192.168.0.16"),
+			expectedGateway: (*ipamv1.IPAddressStr)(ptr.To("192.168.0.1")),
+			expectedPrefix:  24,
+		}),
+		Entry("One pool, with start and existing address, ipAddress annotation present but alraedy acquired", testCaseAllocateAddress{
```
lentzi90, line 1948:
```suggestion
		Entry("One pool, with start and existing address, ipAddress annotation present but already acquired", testCaseAllocateAddress{
```

ipam/ippool_manager_test.go
```diff
@@ -2230,6 +2349,124 @@ var _ = Describe("IPPool manager", func() {
...
+			},
+			addresses: map[ipamv1.IPAddressStr]string{
+				ipamv1.IPAddressStr("192.168.0.12"): "bcde",
+				ipamv1.IPAddressStr("192.168.0.11"): "abcd",
+			},
+			expectedAddress: ipamv1.IPAddressStr("192.168.0.16"),
+			expectedGateway: (*ipamv1.IPAddressStr)(ptr.To("192.168.0.1")),
+			expectedPrefix:  24,
+		}),
+		Entry("One pool, with start and existing address, ipAddress annotation but alraedy acquired", testCapiCaseAllocateAddress{
```
lentzi90, line 2381:
```suggestion
		Entry("One pool, with start and existing address, ipAddress annotation but already acquired", testCapiCaseAllocateAddress{
```

lentzi90 (commented):
Great! Now it looks good to me, just two typos below
/approve
We need to think carefully if we want this in the next release or hold it until after. We already had the beta release so introducing new features at this stage is a bit risky. If it isn't urgent to get this in, I would say it is better to wait for the release-1.10 branch to be created until we merge.
I'm putting hold so we don't accidentally merge it.
/hold

ksahil12:
@lentzi90 actually we were looking forward for this change to get merged. Please check if it's possible to merge it to next release (1.10).
By the way, you must be working on a release branch for 1.10 release, right? In that case, is there any harm in merging this to main?

lentzi90:
We haven't created the release branch yet. As soon as we do, there is no issue merging this to main.
Would it be enough for you to get this on main for now (and then obviously included in v1.11)?

/test metal3-centos-e2e-integration-test-main metal3-ubuntu-e2e-integration-test-main

ksahil12:
I will definitely strive for getting it in 1.10 release :) and would request the same. Please let me know feasibility of it.
Generally, how long are the release cycles (duration between two)?

lentzi90:
/cc @kashifest @adilGhaffarDev 
what do you think? Can we take this in v1.10? I don't see it as a huge risk, but it is a bit late in the cycle.
We should probably think about formalizing code freeze deadlines for future cycles...

lentzi90:
> Generally, how long are the release cycles (duration between two)?

We follow [Cluster API](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/release/release-cycle.md#release-cycle) on this. It is roughly 4 months between minor releases.

adilGhaffarDev:
> what do you think? Can we take this in v1.10? I don't see it as a huge risk, but it is a bit late in the cycle.

Could we add some documentation for this feature, explaining what it does and how users can make use of it?
I am fine with merging this once it is documented.

lentzi90:
Good point! I think the perfect place for this documentation would be here: https://book.metal3.io/ipam/introduction#ipclaim, which is coming from [here](https://github.com/metal3-io/metal3-docs/tree/main/docs/user-guide/src/ipam)

ksahil12:
sharing doc update PR: https://github.com/metal3-io/metal3-docs/pull/515
please review.

ksahil12:
@tuminoid @lentzi90 can you please let me know next steps for the PR?

adilGhaffarDev (commented):
/lgtm
Lets merge this and also the documentation.
@lentzi90 @peppi-lotta

PR #1008, opened 2025-04-24
:seedling: Fix notes.go so that it still complies with forbidigo

peppi-lotta:
**What this PR does / why we need it**: Release notes are not being generated in a file when `log` is used. We need to use `fmt` instead but [forbidigo](https://github.com/ashanbrown/forbidigo) complains about using fmt so I added an ignore to the `.golangci.yaml`.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -179,6 +179,10 @@ issues:
     - stylecheck
     text: "ST1016: methods on the same type should have the same receiver name"
     path: .*(api|types)\/.*\/conversion.*\.go$
+  - path: notes\.go
+    linters:
+    - forbidigo
+    text: use of `fmt.Printf` forbidden by pattern || use of `fmt.Println` forbidden by pattern
   include:
   - EXC0002 # include "missing comments" issues from golangci-lint
   max-issues-per-linter: 0
--- a/hack/tools/release/notes.go
+++ b/hack/tools/release/notes.go
@@ -226,43 +226,43 @@ func run() int {
 		merges[superseded] = append(merges[superseded], "- `<insert superseded bumps and reverts here>`")
 	}
 
-	log.Println("<!-- markdownlint-disable no-inline-html line-length -->")
+	fmt.Println("<!-- markdownlint-disable no-inline-html line-length -->")
 	// TODO Turn this into a link (requires knowing the project name + organization)
-	log.Printf("# Changes since %v\n\n", lastTag)
+	fmt.Printf("# Changes since %v\n\n", lastTag)
 
 	// print the changes by category
 	for _, key := range outputOrder {
 		mergeslice := merges[key]
 		if len(mergeslice) > 0 {
-			log.Printf("## %v\n\n", key)
+			fmt.Printf("## %v\n\n", key)
 			for _, merge := range mergeslice {
-				log.Println(merge)
+				fmt.Println(merge)
 			}
-			log.Println()
+			fmt.Println()
 		}
 
 		// if we're doing beta/rc, print breaking changes and hide the rest of the changes
 		if key == warning {
 			if isBeta(latestTag) {
-				log.Printf(warningTemplate, "BETA RELEASE")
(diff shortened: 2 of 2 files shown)


Sunnatillo (commented):
/lgtm

hack/tools/release/notes.go
```diff
@@ -261,8 +261,8 @@ func run() int {
 		log.Printf("</details>\n\n")
```
smoshiur1237, line 261: do we need to add here also` fmt2`

peppi-lotta: Yes! You are correct.

smoshiur1237 (commented):
/lgtm

hack/tools/release/notes.go
```diff
@@ -22,7 +22,7 @@ package main
 import (
 	"context"
 	"flag"
-	"fmt"
+	fmt2 "fmt"
```
tuminoid, line 25: Why fmt2?

.golangci.yaml
```diff
@@ -179,6 +179,10 @@ issues:
     - stylecheck
     text: "ST1016: methods on the same type should have the same receiver name"
     path: .*(api|types)\/.*\/conversion.*\.go$
+  - path: notes\.go
```
tuminoid, line 182: This could match something else by accident, give it partial path as well.

`release/notes.go` or something.

hack/tools/release/notes.go
peppi-lotta: Forbidigo can't match patterns when the `fmt` package is not imported as the default name so I used fmt2 to avoid forbidigo errors. I realized this is not a good approach and I have now changed the approach to use `fmt` and exclude errors in this file.

.golangci.yaml
peppi-lotta: How should this be written? Using just `release/notes.go` doesn't seem to work.

tuminoid: Looking the paths above, it seems to be regex, so you need to escape `/` -> `\/`.

peppi-lotta: I've tried (among other things)
 - `release\/notes\.go`
 - `.*\/release\/notes\.go`
 - `.*release\/notes\.go`
 
Based on my regex understanding all of the above should but I've only gotten `make test` pass with `notes\.go`.

tuminoid: OK, nevermind. The lint command is in the local directory of release/ already, so it does not appear in the file path at all, which is why none of the paths match.

.golangci.yaml
```diff
@@ -179,6 +179,10 @@ issues:
...
+    linters:
+    - forbidigo
+    text: use of `fmt.Printf` forbidden by pattern || use of `fmt.Prinln` forbidden by pattern
```
tuminoid, line 185:
```suggestion
    text: use of `fmt.Printf` forbidden by pattern || use of `fmt.Println` forbidden by pattern
```
There is a typo here though, and it does not seem to matter, why is that? Replacing the text by garbage does matter, so it does something ...

.golangci.yaml
peppi-lotta: Right! That explains it.

.golangci.yaml
peppi-lotta: I fixed typo.

tuminoid (commented):
/approve

adilGhaffarDev:
/lgtm
@peppi-lotta do we need the same fix in capm3 and BMO too? If yes, please open PRs there too.

tuminoid:
> /lgtm @peppi-lotta do we need the same fix in capm3 and BMO too? If yes, please open PRs there too.

That depends if the linter issues about forbidigo was merged in those repos and/or if linter is even configured to validate the notes.go.

We also might want to actually finalize https://github.com/metal3-io/project-infra/pull/944 to not have fix these in each repo.

adilGhaffarDev:
Changes are only in the release note generation scripts, and these tests don't test that.

PR #1023, opened 2025-04-30
:seedling: Update 1.11 to metadata.yaml

peppi-lotta:
**What this PR does / why we need it**: Update 1.11 to metadata.yaml. This is needed when testing main


lentzi90 (commented):
/approve

Issue #1071, opened 2025-07-17
IPPool controller enters infinite reconciliation loop due to unconditional timestamp updates


ksahil12:
**What steps did you take and what happened:**
Deploy an IPPool resource and observe the controller behavior. The IPPool controller enters an infinite reconciliation loop where it continuously patches the IPPool object with updated `LastUpdated` timestamps, even when no actual IP address allocations or claims require processing.

Steps to reproduce:
1. Create an IPPool resource with any valid configuration
2. Monitor the controller logs - observe repeated "Fetching IPAddress objects" messages
3. Monitor the IPPool resource - observe `status.lastUpdated` field being updated continuously
4. Check API server metrics - observe high PATCH request volume for IPPool objects

**What did you expect to happen:**
The IPPool controller should only update the `status.lastUpdated` field when actual changes occur to IP allocations, not on every reconcile cycle. The controller should reach a steady state with minimal API calls when no address claims need processing.

**Anything else you would like to add:**

**Root Cause Analysis:**
The issue is in ippool_manager.go where `updateStatusTimestamp()` is called unconditionally in multiple methods:
- Line 258: `m3UpdateAddresses()` always calls `updateStatusTimestamp()`
- Line 296: `capiUpdateAddresses()` always calls `updateStatusTimestamp()`  
- Line 886: `deleteAddress()` always calls `updateStatusTimestamp()`
- Line 962: `capiDeleteAddress()` always calls `updateStatusTimestamp()`

**Technical Details:**
Each reconcile cycle calls both `m3UpdateAddresses()` and `capiUpdateAddresses()`, and both methods call `getIndexes()` which performs expensive LIST operations. Even when no claims exist or require processing, the methods still update timestamps, triggering new reconcile cycles.

The `getIndexes()` method already correctly handles timestamp updates (line 195) only when allocations actually change. The additional unconditional calls can be removed.

/kind bug

ksahil12:
Hi @tuminoid , tagging you to get the visibility on this one.

ksahil12:
@tuminoid any timeline when this fix will be released?

PR #1087, opened 2025-08-04
🐛 Fix infinite reconciles and timestamp updates

adilGhaffarDev:
- Patch only when IPPool is changed.
- Removed unnecessary m.updateStatusTimestamp() calls
NOTE: In API it says:
	`// LastUpdated identifies when this status was last observed.`
This seems wrong, firstly I dont think its idea to update something in ippool every time it is observed and secondly even if we are doing it its not named properly. We should change the this text.
- When pool is created for the first time, `IPPool.Status.LastUpdated` is same as `IPPool.CreationTimestamp`. `IPPool.Status.LastUpdated` will only be updated when there is actual change in the ippool which we are doing here: https://github.com/metal3-io/ip-address-manager/blob/main/ipam/ippool_manager.go#L195

Fixes https://github.com/metal3-io/ip-address-manager/issues/1071

### DIFF
--- a/controllers/ippool_controller.go
+++ b/controllers/ippool_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"reflect"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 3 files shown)


controllers/ippool_controller.go
```diff
@@ -83,16 +84,22 @@ func (r *IPPoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
...
 		return ctrl.Result{}, errors.Wrap(err, "failed to init patch helper")
 	}
 	// Always patch ipamv1IPPool exiting this function so we can persist any IPPool changes.
 	defer func() {
-		err = helper.Patch(ctx, ipamv1IPPool)
-		if err != nil {
-			metadataLog.Info("failed to Patch ipamv1IPPool")
-			rerr = err
+		if !reflect.DeepEqual(original, ipamv1IPPool) {
+			err = helper.Patch(ctx, ipamv1IPPool)
```
lentzi90, line 98: Couldn't we add updateStatusTimestamp here?

```suggestion
		if !reflect.DeepEqual(original, ipamv1IPPool) {
		    m.updateStatusTimestamp()
			err = helper.Patch(ctx, ipamv1IPPool)
```

lentzi90 (commented):
Do I understand correctly that this would make the `lastUpdated` field basically the same as the creation timestamp?
I have one idea below how to maybe improve it a bit

PR #1088, opened 2025-08-04
🌱 Bump osv-scanner to v2.1.0

kashifest:
- Bumps in the verify-release.sh
- Bumps in the osv-scanner workflow


smoshiur1237 (commented):
/lgtm

PR #1209, opened 2025-12-01
:seedling: WIP: Add base for e2e tests

peppi-lotta:
This PR adds the base for e2e test:

- GitHub Actions workflow for e2e testing
- Test suite setup
- Support for multiple BMO (Bare Metal Operator) deployment releases (0.9-0.11)
- Ironic standalone operator configurations with TLS and basic auth components
- Helper scripts for CI execution, environment setup, and cleanup
- Integration with cert-manager and sushy-tools for BMC simulation

**Checklist:**


lentzi90:
You are absolutely correct. What I am after is to do it one step at a time. We are already working on getting something like this in CAPM3, which we could then reuse here with less work needed.

Regardless of this, having a light weight test suite that still tests IPAM in a real cluster would be valuable. Just like we have the fixture tests in BMO. It can run in 5-10 minutes easily and would work well on any normal laptop. It also makes it way easier to test corner cases without the need for heavy VMs. For example, we could easily scale test IPAM with this. Scale testing through CAPM3 is much harder and either requires lots of resources or faking.

In any case, I am not against having a complete e2e suite here that includes CAPM3. I would suggest getting it to work in CAPM3 first though, since CAPM3 can work without IPAM. Another option would be to target a different infrastructure provider. In theory, IPAM should work with CAPD or CAPIM for example. That would be interesting to see!

PR #1259, opened 2026-01-09
🌱 rename lint-full -> lint

tuminoid:
Our github workflow for golangci-lint matches lint-full, so its quite confusing and not really productive to separate between full and fast, as the difference is like 1s.


tuminoid:

lentzi90 (commented):
/approve

PR #1290, opened 2026-02-04
🌱  Add basic fuzz test

smoshiur1237:
Adding basic Fuzz test for IPAM. It can be a baseline for Fuzz test. Also added necessary make target to have proper check.


test/fuzz/utils_fuzzer_test.go
tuminoid: The input are too small to test anything, and testing a testing suite is not adding any value. They'll get tested during the fuzz tests and proper inputs. Drop the whole file.

test/fuzz/README.md
```diff
@@ -53,3 +59,50 @@
...
+name in the commands above.
+
+## Example: Running the Utils Fuzzer
+
+To run the new `FuzzFilterAndContains` fuzzer:
+
+```bash
+mkdir -p output
+cd output
+go114-fuzz-build -o utils_FuzzFilterAndContains.a -func FuzzFilterAndContains ../test/fuzz/
```
tuminoid: `go test -fuzz ...` is the modern syntax.

tuminoid (commented):
/hold

This needs to be rewritten in modern Go fuzzing tools and syntaxes, and is also missing all scaffolding - make targets, workflow integrations etc.

adilGhaffarDev:
I was reading CAPI docs:

https://cluster-api.sigs.k8s.io/developer/core/testing?utm_source=chatgpt.com#fuzzing-tests 

In CAPI they have api conversion fuzzing in the project and other fuzz tests are in the cncf-fuzzing project: https://github.com/cncf/cncf-fuzzing/tree/main/projects/cluster-api .

In ipam we dont need conversion fuzzing because there is no new api yet. In capm3 I have added it for the conversion(https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/api/v1beta1/conversion_test.go).
I would suggest we also use cncf-fuzzing project.

I think we should also need to decide what we want to test in our fuzz tests. Probably not good idea to test each and every function.

Makefile
```diff
@@ -96,6 +96,26 @@ unit-cover: ## Run unit tests with code coverage
...
 	cd ..
 
+.PHONY: fuzz
+fuzz: ## Run fuzz tests with seed corpus (no fuzzing, regression test only)
+	cd test/fuzz && go test -v ./...
+
+.PHONY: fuzz-run
+fuzz-run: ## Run fuzz tests with fuzzing enabled (default: 30s per test)
+	@echo "Running fuzz tests for 30 seconds each..."
+	cd test/fuzz && go test -fuzz=FuzzNewIPPoolManager -fuzztime=30s
```
tuminoid: Make FUZZ_TIME= configurable, then we can only have one fuzz target.

smoshiur1237 (commented):
@adilGhaffarDev @Rozzii @lentzi90 
Is it okay to take this in? I have addressed all the reviews. I will add any new tests if necessary in the followup PR.

Rozzii (commented):
/approve
Yeah why not, it does not seem to cause any modifications to any other components or tests other than minimal changes to go.mod and sum in the /test

smoshiur1237 (commented):
/test metal3-centos-e2e-integration-test-main
/test metal3-ubuntu-e2e-integration-test-main

PR #1294, opened 2026-02-05
:bug: Revert listing limits

fmuyassarov:
It seems that limiting the listing to 200 wasn't the best number, specifically for cases when there is more objects than 200 on the same namespace. Revert it for now and instead we improve the filtering the objects during fetch.

Related to https://github.com/metal3-io/cluster-api-provider-metal3/issues/2924.


lentzi90 (commented):
/lgtm

PR #1359, opened 2026-04-06
✨ Add random IP allocation strategy to IPPool

shibaPuppy:
Introduce an AllocationStrategy field to IPPoolSpec that allows choosing between "sequential" (default) and "random" allocation.

When set to "random", a random offset within the pool is chosen and the pool is scanned forward (wrapping around at the end) until the first available IP is found. This walks each index at most once, giving O(1) memory usage.

- Add AllocationStrategy type with "sequential"/"random" enum values
- Extract selectIPFromPool() helper to share logic between allocateAddress() and  capiAllocateAddress()
- Add webhook defaulting (sequential) and immutability validation
- PreAllocations and requestedIP annotation bypass random strategy

Currently, IP addresses are always allocated sequentially (first available IP from the pool). This PR introduces an
 allocationStrategy field to IPPoolSpec that allows choosing between sequential (default, existing behavior) and   
random allocation.

When set to random, a random offset within the pool is chosen and the pool is scanned forward (wrapping around at
the end) until the first available IP is found. Each index is visited at most once, so allocation uses O(1) memory.

Key changes:
- Add AllocationStrategy type with sequential/random enum values to IPPoolSpec
- Extract selectIPFromPool() helper to share allocation logic between allocateAddress() and capiAllocateAddress()
- Add webhook defaulting (sequential) and immutability validation
- PreAllocations and ipAddress annotation bypass random strategy and always use sequential scan

                                                                                                                     
```yaml
kind: IPPool                                                                                                       
spec:                                                                                                              
  allocationStrategy: random
  pools:                                                                                                           
    - start: 192.168.0.10                                 
      end: 192.168.0.50  
      prefix: 24                                                                                                   
      gateway: 192.168.0.1
```

Fixes #

**Checklist:**

### DIFF
--- a/api/v1alpha1/ippool_types.go
+++ b/api/v1alpha1/ippool_types.go
@@ -26,6 +26,16 @@ const (
 	IPPoolFinalizer = "ippool.ipam.metal3.io"
 )
 
+// AllocationStrategy defines the strategy for IP address allocation from a pool.
+type AllocationStrategy string
+
+const (
+	// AllocationStrategySequential allocates IPs sequentially from pools (default).
+	AllocationStrategySequential AllocationStrategy = "sequential"
+	// AllocationStrategyRandom allocates IPs randomly from available pool addresses.
+	AllocationStrategyRandom AllocationStrategy = "random"
+)
+
 // MetaDataIPAddress contains the info to render th ip address. It is IP-version
 // agnostic.
 type Pool struct {
@@ -61,6 +71,16 @@ type IPPoolSpec struct {
 
 	// Pools contains the list of IP addresses pools
 	Pools []Pool `json:"pools,omitempty"`
+
+	// +kubebuilder:default=sequential
+	// +kubebuilder:validation:Enum=sequential;random
+	// AllocationStrategy defines how IP addresses are allocated from the pools.
+	// "sequential" (default) allocates the first available IP.
+	// "random" allocates a random available IP.
+	// In both strategies, multiple pools are consumed in declaration order: a
+	// pool is fully exhausted before the next one is used, and the strategy only
+	// changes how an address is selected within a single pool.
+	AllocationStrategy AllocationStrategy `json:"allocationStrategy,omitempty"`
 
 	// PreAllocations contains the preallocated IP addresses
 	PreAllocations map[string]IPAddressStr `json:"preAllocations,omitempty"`
--- a/api/v1alpha1/utils.go
+++ b/api/v1alpha1/utils.go
@@ -19,6 +19,7 @@ package v1alpha1
 import (
 	"errors"
 	"fmt"
+	"math"
 	"math/big"
 	"net"
 )
@@ -74,6 +75,106 @@ func GetIPAddress(entry Pool, index int) (IPAddressStr, error) {
 		}
 	}
 	return IPAddressStr(ip.String()), nil
+}
+
+// ValidatePool validates the address-range fields of a single pool entry: that
+// any Start/End/Subnet values present are well-formed, and that start <= end
+// when both Start and End are set. It is intentionally lenient about presence
+// (a Start-only or End-only pool is valid for the sequential strategy), so it
+// does not require Start or Subnet to be set and does not bound-check the size.
+// It is the shared validation used by both the webhook and GetPoolSize so the
+// range/format checks live in a single place.
+func ValidatePool(entry Pool) error {
+	var startIP, endIP net.IP
+
+	if entry.Start != nil {
+		startIP = net.ParseIP(string(*entry.Start))
+		if startIP == nil {
+			return fmt.Errorf("invalid Start IP %q", *entry.Start)
+		}
+	}
+	if entry.End != nil {
+		endIP = net.ParseIP(string(*entry.End))
+		if endIP == nil {
+			return fmt.Errorf("invalid End IP %q", *entry.End)
+		}
+	}
+	if entry.Subnet != nil {
+		if _, _, err := net.ParseCIDR(string(*entry.Subnet)); err != nil {
+			return fmt.Errorf("invalid Subnet %q: %w", *entry.Subnet, err)
+		}
+	}
+	if startIP != nil && endIP != nil {
+		if new(big.Int).SetBytes(startIP.To16()).Cmp(new(big.Int).SetBytes(endIP.To16())) > 0 {
+			return fmt.Errorf("end IP %s is before start IP %s", endIP, startIP)
+		}
+	}
+	return nil
+}
+
+// GetPoolSize returns the number of indexable IP addresses in the given pool
+// entry, matching the index space accepted by GetIPAddress: GetIPAddress(entry, i)
+// is valid for i in [0, size) and returns an error for i >= size. Unlike
+// ValidatePool, it requires the pool to be bounded (Start with End/Subnet, or
+// Subnet only) so that a finite size can be computed.
+func GetPoolSize(entry Pool) (int, error) {
+	if err := ValidatePool(entry); err != nil {
+		return 0, err
+	}
+	if entry.Start == nil && entry.Subnet == nil {
+		return 0, errors.New("either Start or Subnet is required for ipAddress")
+	}
+
+	var startIP, endIP net.IP
+	inclusiveStart := true
+
+	switch {
+	case entry.Start != nil && entry.End != nil:
+		startIP = net.ParseIP(string(*entry.Start))
+		endIP = net.ParseIP(string(*entry.End))
+	case entry.Start != nil && entry.Subnet != nil:
+		_, ipNet, err := net.ParseCIDR(string(*entry.Subnet))
+		if err != nil {
+			return 0, err
+		}
+		startIP = net.ParseIP(string(*entry.Start))
+		endIP = lastIPInSubnet(ipNet)
+	case entry.Start != nil:
+		return 0, errors.New("pool with Start requires End or Subnet to determine size")
+	default:
+		_, ipNet, err := net.ParseCIDR(string(*entry.Subnet))
+		if err != nil {
+			return 0, err
+		}
+		// GetIPAddress with Subnet-only maps index 0 to network+1, so the
+		// network address itself is excluded from the index space.
+		startIP = ipNet.IP
+		endIP = lastIPInSubnet(ipNet)
+		inclusiveStart = false
+	}
+
+	s := new(big.Int).SetBytes(startIP.To16())
+	e := new(big.Int).SetBytes(endIP.To16())
+	diff := new(big.Int).Sub(e, s)
+	if diff.Sign() < 0 {
+		return 0, fmt.Errorf("end IP %s is before start IP %s", endIP, startIP)
+	}
+	if inclusiveStart {
+		diff.Add(diff, big.NewInt(1))
+	}
+	if !diff.IsInt64() || diff.Int64() > math.MaxInt {
+		return 0, errors.New("pool size exceeds int range")
+	}
+	return int(diff.Int64()), nil
+}
+
+// lastIPInSubnet returns the highest address contained in the given CIDR.
+func lastIPInSubnet(n *net.IPNet) net.IP {
+	last := make(net.IP, len(n.IP))
+	for i := range n.IP {
+		last[i] = n.IP[i] | ^n.Mask[i]
+	}
+	return last
 }
 
 // addOffsetToIP computes the value of the IP address with the offset. It is
--- a/ipam/ippool_manager.go
+++ b/ipam/ippool_manager.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand/v2"
 	"net"
 	"reflect"
 	"strings"
@@ -453,38 +454,37 @@ func (m *IPPoolManager) allocateAddress(addressClaim *ipamv1.IPClaim,
 		if ipAllocated {
 			break
 		}
-		index := 0
-		for !ipAllocated {
-			allocatedAddress, err = ipamv1.GetIPAddress(pool, index)
-			if err != nil {
-				break
-			}
-			index++
-			// Check if requestedIP is present and matches the current address
-			if requestedIP != "" && !m.ipEqual(allocatedAddress, requestedIP) {
-				continue
-			}
-			if requestedIP != "" && m.ipEqual(allocatedAddress, requestedIP) {
-				isRequestedIPAllocated = true
-			}
-			// We have a pre-allocated ip, we just need to ensure that it matches the current address
-			// if it does not, continue and try the next address
-			if ipPreAllocated && !m.ipEqual(allocatedAddress, preAllocatedAddress) {
-				continue
-			}
-			// Here the two addresses match, so we continue with that one
-			if ipPreAllocated {
-				ipAllocated = true
-			}
-			// If we have a preallocated address, this is useless, otherwise, check if the
-			// ip is free
-			if _, ok := addresses[allocatedAddress]; !ok && allocatedAddress != "" {
-				ipAllocated = true
-			}
-			if !ipAllocated {
-				continue
-			}
 
+		// PreAllocations and requestedIP always use sequential scan
+		if ipPreAllocated || requestedIP != "" {
+			index := 0
+			for !ipAllocated {
+				allocatedAddress, err = ipamv1.GetIPAddress(pool, index)
+				if err != nil {
+					break
+				}
+				index++
+				if requestedIP != "" && !m.ipEqual(allocatedAddress, requestedIP) {
+					continue
+				}
+				if requestedIP != "" && m.ipEqual(allocatedAddress, requestedIP) {
+					isRequestedIPAllocated = true
+				}
+				if ipPreAllocated && !m.ipEqual(allocatedAddress, preAllocatedAddress) {
+					continue
+				}
+				if ipPreAllocated {
+					ipAllocated = true
+				}
+				if _, ok := addresses[allocatedAddress]; !ok && allocatedAddress != "" {
+					ipAllocated = true
+				}
+			}
+		} else {
+			allocatedAddress, ipAllocated = m.selectIPFromPool(pool, addresses)
+		}
+
+		if ipAllocated {
 			if pool.Prefix != 0 {
 				prefix = pool.Prefix
 			}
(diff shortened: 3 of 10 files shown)


tuminoid (commented):
/ok-to-test
/cc @peppi-lotta

Sunnatillo:
This is good feature. But could you explain your use case more. 

IMHO,  this kind of changes should come with proposal or at least documentation. 
I will wait others opinions.

api/v1alpha1/utils.go
```diff
@@ -76,6 +76,102 @@ func GetIPAddress(entry Pool, index int) (IPAddressStr, error) {
...
+			size := big.NewInt(0).Add(diff, big.NewInt(1))
+			if !size.IsInt64() || size.Int64() > int64(int(^uint(0)>>1)) {
+				return int(^uint(0) >> 1), nil
+			}
+			return int(size.Int64()), nil
+		}
+
+		// Start only, no End, no Subnet: cannot determine size
+		return 0, errors.New("cannot determine pool size without End or Subnet")
+	}
```
peppi-lotta, line 105: We have `validatePoolRanges` function in the webhooks.
https://github.com/metal3-io/ip-address-manager/blob/958ca5407dcac5966786a24598ec65d1523f1af7/internal/webhooks/v1alpha1/ippool_webhook.go#L204

I think it would be better to move that function in to the `api/v1alpha1/ippool_types.go` and call it here (and in the webhook validation function)  instead of rewriting the whole logic here. 

The `validatePoolRanges` function is missing the Start + Subnet (no End) validation so you could add that in there.

shibaPuppy:
@Sunnatillo 
https://github.com/metal3-io/metal3-docs/pull/679 add design docs

api/v1alpha1/utils.go
shibaPuppy: @peppi-lotta 
thanks for the review! I've refactored to share the validation logic.

- Moved validateIP to api/v1alpha1/ippool_types.go as ValidateIPAddress() — now shared across all webhook files (ippool, ipaddress, ipclaim)
- Added ValidatePool() in the same package for per-pool validation (IP format, CIDR format, start <= end)
- Webhook's validatePoolRanges() now calls ipamv1.ValidateIPAddress() for IP validation while keeping per-field error paths (spec.pools[i].start, etc.) for precise error reporting
- GetPoolSize() assumes the pool has already been validated by the webhook via ValidatePool(), so it focuses only on size computation

Sunnatillo:
@shibaPuppy Please rebase the PR. We will try to take this in this release. Today or tomorrow

peppi-lotta:
Please also squash your commits into one

api/v1alpha1/utils.go
```diff
@@ -76,6 +76,93 @@ func GetIPAddress(entry Pool, index int) (IPAddressStr, error) {
 	return IPAddressStr(ip.String()), nil
 }
 
+// GetPoolSize returns the total number of allocatable IP addresses in a Pool entry.
+// It assumes the Pool has already been validated by the webhook via ValidatePool().
+func GetPoolSize(entry Pool) (int, error) {
```
peppi-lotta: This function is defined and tested but I don't see where it is actually used. Please remove if it is not needed

ipam/ippool_manager.go
```diff
@@ -985,6 +984,48 @@ func (m *IPPoolManager) capiDeleteAddress(ctx context.Context,
...
+const maxRandomCandidates = 1000
+
+// selectIPFromPool picks an available IP from the given pool based on the allocation strategy.
+// Returns the selected IP and true, or empty string and false if no IP is available.
+func (m *IPPoolManager) selectIPFromPool(pool ipamv1.Pool, addresses map[ipamv1.IPAddressStr]string) (ipamv1.IPAddressStr, bool) {
+	if m.IPPool.Spec.AllocationStrategy == ipamv1.AllocationStrategyRandom {
+		// Collect available IPs first (pool IPs minus already-allocated),
+		// capped at maxRandomCandidates to bound memory usage.
+		available := make([]ipamv1.IPAddressStr, 0, maxRandomCandidates)
+		for index := 0; len(available) < maxRandomCandidates; index++ {
```
peppi-lotta, line 1009: This selects first 1000 available IPs and allocates randomly from them. If we have a pool of 2000 ips and no ips are allocated yet then it will select randomply from the first 1000 ips and ignore the 1001-2000 range. I'm ok with that but it would be nice to have this documented.

ipam/ippool_manager_test.go
```diff
@@ -3079,4 +3209,305 @@ var _ = Describe("IPPool manager", func() {
...
+			if ip[i] < start[i] {
+				return false
+			}
+			if ip[i] > end[i] {
+				return false
+			}
+		}
+		return true
+	}
+
```
peppi-lotta, line 3228: Please add test case where byte boundaries are crossed. This helper function is not working correctly and this fails. You can run https://go.dev/play/p/KclI9UbRCyL this play ground for a quick test. 
```suggestion
	Context("isIPInRange helper", func() {
		It("should handle ranges that cross byte boundaries", func() {
			// 192.168.1.5 is inside [192.168.0.10, 192.168.2.20] but the
			// per-byte comparison incorrectly rejects it because byte 3
			// (5) < start byte 3 (10).
			Expect(isIPInRange("192.168.1.5", "192.168.0.10", "192.168.2.20")).To(BeTrue(),
				"192.168.1.5 should be within 192.168.0.10-192.168.2.20")
		})
	})
```

ipam/ippool_manager.go
shibaPuppy: @peppi-lotta
I expanded the maxRandomCandidates doc comment to spell out the implication with a concrete

api/v1alpha1/utils.go
shibaPuppy: @peppi-lotta
it along with its helpers (ipDiff, capToMaxInt, lastIPInSubnet) and the associated DescribeTable entries in utils_test.go

ipam/ippool_manager_test.go
shibaPuppy: @peppi-lotta 
fixed the helper to use bytes.Compare on To16(), and added your suggested test plus boundary-inclusive / outside-range cases under a new context

peppi-lotta:
@tuminoid can you trigger co-pilot review here?

tuminoid:
> @tuminoid can you trigger co-pilot review here?

Done.

shibaPuppy:
@peppi-lotta @Sunnatillo @tuminoid 
all four points are addressed:
                                                                                             
1. **enum marker duplication** (`api/v1alpha1/ippool_types.go`) — Removed the duplicate    
  `+kubebuilder:validation:Enum` marker from the type (kept on the field to match this repo's
   convention). Regenerated the CRD; the `allOf` block collapsed into a single `enum`.       
                                                            
2. **flaky non-sequential test** (`ipam/ippool_manager_test.go`) — Replaced the            
  probabilistic assertion with two deterministic cases: one verifies the random branch
  returns the only remaining IP when others are pre-allocated, the other verifies the        
  exhausted-pool error path.                                

3. **allocationStrategy immutability** (`internal/webhooks/v1alpha1/ippool_webhook.go`) —  
  Tightened to true immutability. Both sides are now normalized to `sequential` when empty
  before comparison, so legacy objects cannot switch to random on update.                    
                                                            
4. **validatePoolRanges doc/impl mismatch**                                                
  (`internal/webhooks/v1alpha1/ippool_webhook.go`) — Applied your suggested comment. Kept the
   inline per-field validation since it preserves precise `field.ErrorList` paths (e.g.,     
  `spec.pools[i].start`) for error reporting.

ipam/ippool_manager.go
tuminoid: So if pool is nearly exhausted, it could do pool-1 loops. Sequential does the same, but uff. A near exhausted /16 could loop 64k times.

ipam/ippool_manager_test.go
```diff
@@ -3079,4 +3210,340 @@ var _ = Describe("IPPool manager", func() {
...
+			addresses := map[ipamv1.IPAddressStr]string{}
+			ipPoolMgr, err := NewIPPoolManager(nil, ipPool, logr.Discard())
+			Expect(err).NotTo(HaveOccurred())
+
+			// Allocate all 11 IPs and verify no duplicates
+			for i := range 11 {
+				ipClaim := &ipamv1.IPClaim{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "TestRef-" + string(rune('a'+i)),
```
tuminoid: Since it ranges only up to 11, it produces ascii, but if that range goes beyond 26, then its non-ascii generator. At least add a comment about this?

api/v1alpha1/ippool_types.go
```diff
@@ -118,3 +139,58 @@ type IPPoolList struct {
...
+		return nil
+	}
+	if net.ParseIP(string(s)) == nil {
+		return fmt.Errorf("invalid IP address: %q", s)
+	}
+	return nil
+}
+
+// ValidatePool validates a single Pool entry: IP format, CIDR format, and start <= end.
+func ValidatePool(pool Pool) []string {
```
tuminoid: Dead code, nobody calls this?

tuminoid (commented):
Looking much better!

ipam/ippool_manager.go
```diff
@@ -985,6 +984,58 @@ func (m *IPPoolManager) capiDeleteAddress(ctx context.Context,
...
+// the first 1000 are freed.
+const maxRandomCandidates = 1000
+
+// selectIPFromPool picks an available IP from the given pool based on the allocation strategy.
+// Returns the selected IP and true, or empty string and false if no IP is available.
+func (m *IPPoolManager) selectIPFromPool(pool ipamv1.Pool, addresses map[ipamv1.IPAddressStr]string) (ipamv1.IPAddressStr, bool) {
+	if m.IPPool.Spec.AllocationStrategy == ipamv1.AllocationStrategyRandom {
+		// Collect up to maxRandomCandidates available IPs starting from the
+		// beginning of the pool, then pick one uniformly at random. See the
+		// maxRandomCandidates doc for the implications on pools larger than the cap.
```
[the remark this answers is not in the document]
shibaPuppy: updated the PR description — those utilities (GetPoolSize, lastIPInSubnet) were dropped in an earlier review iteration as unused, so the description is now in sync with the implementation.

ipam/ippool_manager.go
shibaPuppy: @tuminoid 
i agreed — sequential has the same worst-case shape today, so random isn't strictly worse here, just more consistent about hitting it. 
given the expected IPPool sizing for metal3 (roughly /24~/20), the scan stays in the few-thousand-iteration range at worst, which is fine. I'm leaving this as-is for now and treating an explicit scan cap as a measured follow-up if anyone with a /16-class pool reports it. 
happy to revisit if you'd rather have it in this PR.

api/v1alpha1/ippool_types.go
shibaPuppy: @tuminoid 
thx, 
ValidatePool` was only used by `GetPoolSize` which was dropped in an earlier review iteration, so the function had no remaining callers.

ipam/ippool_manager_test.go
shibaPuppy: switched to `fmt.Sprintf("TestRef-%d", i)` — that removes the ASCII assumption entirely, so any future loop count stays safe.

ipam/ippool_manager.go
tuminoid: Could you make an issue about it, maybe we'll remember to address it one day.

tuminoid (commented):
LGTM

/cc @peppi-lotta @lentzi90 @adilGhaffarDev

api/v1alpha1/ippool_types.go
```diff
@@ -118,3 +138,14 @@ type IPPoolList struct {
 func init() {
 	objectTypes = append(objectTypes, &IPPool{}, &IPPoolList{})
 }
+
+// ValidateIPAddress validates that the given string is a valid IP address.
+func ValidateIPAddress(s IPAddressStr) error {
```
adilGhaffarDev: why are we moving this here and removing from webhooks?

ipam/ippool_manager.go
adilGhaffarDev: why not start with random? pick random offset in the pool and move forward until you find a free ip. 
Also wrap around if it reaches the end.

You will not need maxRandomCandidates .

api/v1alpha1/ippool_types.go
shibaPuppy: in an earlier iteration of this PR I introduced a `GetPoolSize()` helper (used by the random allocation path) and moved `ValidateIPAddress` / `ValidatePool` out of the webhooks package so both the webhook and `GetPoolSize` could share the same validation. During the review rounds, `GetPoolSize` was dropped (the window-based random implementation didn't need it), and `ValidatePool` lost its only caller and was removed along with it.

what's left is `ValidateIPAddress`, and all of its remaining callers live inside `internal/webhooks/v1alpha1/` (ipaddress / ippool / ipclaim webhooks). So strictly speaking the original justification for the move is gone.

1. **Move it back** to the webhooks package (or inline it as an unexported helper there), since there is no concrete consumer outside webhooks right now.
2. **Keep it in `api/v1alpha1`** — your other comment on the random algorithm proposes a `rand.IntN(poolSize)`-style approach, which would bring `GetPoolSize` back, and that helper would be a legitimate non-webhook caller of `ValidateIPAddress` (and would itself need to live with the API types).

if we adopt the simpler random algorithm you suggested below, option 2 becomes consistent and `ValidateIPAddress` staying in `api/v1alpha1` is justified by a real consumer. If we keep the current window-based approach, I'm happy to move `ValidateIPAddress` back into webhooks in this PR.

shibaPuppy: @adilGhaffarDev 
i really like this — it's simpler, has O(1) memory, and incidentally fixes the asymmetry @peppi-lotta flagged earlier (with the window approach, indices beyond `maxRandomCandidates` in a large pool are never reachable until the first 1000 fill up). quick sketch of what I'd change:

```go
func (m *IPPoolManager) selectIPFromPool(pool ipamv1.Pool, addresses map[ipamv1.IPAddressStr]string) (ipamv1.IPAddressStr, bool) {
    if m.IPPool.Spec.AllocationStrategy == ipamv1.AllocationStrategyRandom {
        size, err := ipamv1.GetPoolSize(pool)
        if err != nil || size <= 0 {
            return "", false
        }
        start := rand.IntN(size) //nolint:gosec
        for i := 0; i < size; i++ {
            index := (start + i) % size
            candidate, err := ipamv1.GetIPAddress(pool, index)
            if err != nil {
                continue
            }
            if _, ok := addresses[candidate]; !ok && candidate != "" {
                return candidate, true
            }
        }
        return "", false
    }
}
```

this PR has been through quite a few review rounds already, so my preference would be to close out the current implementation as-is and tackle the offset+wrap-around rewrite in a follow-up PR.
that keeps the diff here focused on shipping the strategy field and the default behavior, and lets the algorithm swap stand on its own with dedicated tests and benchmarks.
would that work for you, or would you rather we block this PR until the new algorithm is in?

adilGhaffarDev: I would suggest updating this pr, this is not something critical that we want to merge asap. 
Above code is already looking good, so I dont think it will need lot of iterations.

shibaPuppy: @adilGhaffarDev 
I've pushed the review feedback as a separate commit (e18a610) on top of the original change so the delta is easy to review against the previous round. If it looks good to you, I'll squash it into the original commit before merge.

cc. @tuminoid

api/v1alpha1/utils.go
```diff
@@ -76,6 +77,73 @@ func GetIPAddress(entry Pool, index int) (IPAddressStr, error) {
 	return IPAddressStr(ip.String()), nil
 }
 
+// GetPoolSize returns the number of indexable IP addresses in the given pool
+// entry, matching the index space accepted by GetIPAddress: GetIPAddress(entry, i)
+// is valid for i in [0, size) and returns an error for i >= size.
+func GetPoolSize(entry Pool) (int, error) {
```
adilGhaffarDev, line 120: can you add unit tests for this in `utils_test.go`

adilGhaffarDev (commented):
looks good, please add unit test for GetPoolSize

api/v1alpha1/utils.go
shibaPuppy: @adilGhaffarDev 
added unit tests for `GetPoolSize` in utils_test.go.

adilGhaffarDev (approved):
/approve

ipam/ippool_manager.go
```diff
@@ -985,6 +984,47 @@ func (m *IPPoolManager) capiDeleteAddress(ctx context.Context,
...
+			candidate, err := ipamv1.GetIPAddress(pool, index)
+			if err != nil {
+				continue
+			}
+			if _, ok := addresses[candidate]; !ok && candidate != "" {
+				return candidate, true
+			}
+		}
+		return "", false
+	}
```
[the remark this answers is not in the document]
shibaPuppy: @tuminoid 
updated the PR description to match the implementation

ipam/ippool_manager.go
```diff
@@ -985,6 +984,47 @@ func (m *IPPoolManager) capiDeleteAddress(ctx context.Context,
...
+func (m *IPPoolManager) selectIPFromPool(pool ipamv1.Pool, addresses map[ipamv1.IPAddressStr]string) (ipamv1.IPAddressStr, bool) {
+	if m.IPPool.Spec.AllocationStrategy == ipamv1.AllocationStrategyRandom {
+		// Pick a random offset in the pool and scan forward, wrapping around
+		// at the end. This walks each index at most once and returns the first
+		// free IP encountered, giving O(1) memory and removing the asymmetry
+		// of a fixed-size candidate window.
+		size, err := ipamv1.GetPoolSize(pool)
+		if err != nil || size <= 0 {
+			return "", false
+		}
```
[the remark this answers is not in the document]
shibaPuppy: @tuminoid 
random allocation needs a finite, known size to pick a random offset, so `GetPoolSize` erroring on unbounded/oversized pools is correct behavior. 
updated the webhook to reject these only when the strategy is random.

Sunnatillo:
We are having some CI issues. We are fixing it

ipam/ippool_manager.go
peppi-lotta, line 551: Just one more thing to consider: User can specify multiple pools in one IPPool  CR. Is this a conscious decisions to use the pools sequentially? I'm happy to take this is as is but I just want to make sure this will be mentioned somewhere tha even with random allocation the pools are used/exhausted sequentially.

peppi-lotta (commented):
I still think this comment is valid: https://github.com/metal3-io/ip-address-manager/pull/1359/changes#r3051064610

In `GetPoolSize()` you are validating pool range and webhook already has a function for doing that. That function should be made public and used instead of recreating validation in `GetPoolSize()` .  I see you are doing updates to the `validatePoolRanges` function so why not also make it public and use it to do the validation in `GetPoolSize()` instead rewriting?

ipam/ippool_manager.go
shibaPuppy: i read this as the same point from your earlier review. one constraint on the direction, though: `validatePoolRanges` lives in `internal/webhooks/v1alpha1`, while `GetPoolSize` lives in `api/v1alpha1`, and the dependency only goes webhook → api. so `api` can't import `internal/webhooks`, which means `GetPoolSize` can't call the webhook function.

so I'll do it in the direction the dependency allows — put the validation in `api/v1alpha1` and have both sides use it:

- add an exported `ValidatePool(entry Pool) error` in `api/v1alpha1` covering presence + IP/CIDR format + `start <= end`.
- factor the bounds parsing into one internal helper so `GetPoolSize` computes the size from the same parse, which removes the rewritten validation inside `GetPoolSize`.
- change the webhook's `validatePoolRanges` to call `ipamv1.ValidatePool()` instead of validating inline.

this also gives the validation's home in `api/v1alpha1` a concrete in-package consumer, which addresses @adilGhaffarDev's earlier question about why it lived there. Would this direction work?

peppi-lotta: I think the `validatePoolRanges` should be moved out from webhooks and into the `api/v1alpha1` so that it can be reused and we don't have to rewrite the logic. I also think it's ok to the rename and add to the validation code if you think it's missing something that would be important for validation.

shibaPuppy: @peppi-lotta 
added ValidatePool(entry Pool) error in api/v1alpha1, and now both the webhook's validatePoolRanges and GetPoolSize call it instead of each re-implementing the range/format checks.

adilGhaffarDev:
> Just one more thing to consider: User can specify multiple pools in one IPPool CR. Is this a conscious decisions to use the pools sequentially? I'm happy to take this is as is but I just want to make sure this will be mentioned somewhere tha even with random allocation the pools are used/exhausted sequentially.

@shibaPuppy can you address this comment.

shibaPuppy:
> > Just one more thing to consider: User can specify multiple pools in one IPPool CR. Is this a conscious decisions to use the pools sequentially? I'm happy to take this is as is but I just want to make sure this will be mentioned somewhere tha even with random allocation the pools are used/exhausted sequentially.

@adilGhaffarDev 

added it to the `AllocationStrategy` field godoc ([`api/v1alpha1/ippool_types.go#L77-L83`](https://github.com/metal3-io/ip-address-manager/blob/3aed43bd5ed4cbc7445ab78090b985aec8ebc220/api/v1alpha1/ippool_types.go#L77-L83)), and it propagates to the CRD field description via `make generate`:

```go
	// AllocationStrategy defines how IP addresses are allocated from the pools.
	// "sequential" (default) allocates the first available IP.
	// "random" allocates a random available IP.
	// In both strategies, multiple pools are consumed in declaration order: a
	// pool is fully exhausted before the next one is used, and the strategy only
	// changes how an address is selected within a single pool.
	AllocationStrategy AllocationStrategy `json:"allocationStrategy,omitempty"`
```

let me know if you'd prefer it mentioned somewhere else as well (e.g. the book docs).

PR #1410, opened 2026-05-12
:seedling: Add e2e test for IPAM

peppi-lotta:
## Add E2E test framework for IPAM

This PR introduces a standalone end-to-end test suite for the IP Address Manager, built on top of the CAPI test framework with Ginkgo/Gomega.

### What's included

**E2E test infrastructure (`test/e2e/`)**
- Suite setup: bootstraps a Kind cluster, installs CAPI + IPAM providers via clusterctl
- Config and provider metadata for CAPI and CAPD v1.13
- CI script (`hack/e2e/ci-script.sh`) with label-based test filtering using `GINKGO_FOCUS`
- Helper scripts to ensure Go and kubectl meet minimum versions

**Basic tests** (label: `basic`)
- IPPool creation and verification
- IP allocation via Metal3 IPClaim and CAPI IPAddressClaim
- Garbage collection of IPAddress objects when claims are deleted

**Feature tests** (label: `features`)
- Preallocations
- Annotation-based IP requests (Metal3 and CAPI)
- Pool exhaustion and recovery
- Multiple pools in a single IPPool spec
- Pool-level gateway/DNS overrides
- IPPool status tracking
- Mixed Metal3 + CAPI claims on the same pool
- IPPool deletion blocked by finalizer while claims exist

**Documentation**
- Added `docs/e2e.md` covering prerequisites, running tests, configuration variables, CI usage, and test structure

### How to run

```sh
# All tests
.hack/e2e/ci-script.sh

# Basic only
GINKGO_FOCUS_LABELS=basic .hack/e2e/ci-script.sh

# Features only
GINKGO_FOCUS_LABELS=features .hack/e2e/ci-script.sh
```

**Checklist:**

- [x] E2E tests have been added, if necessary.


.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,52 @@
...
+        persist-credentials: false
+
+    - name: Run IPAM e2e Tests
+      env:
+        GINKGO_FOCUS: "${{ inputs.ginkgo-focus }}"
+        E2E_TESTS: "${{ inputs.e2e-tests }}"
+      # We need a new shell to pick up the new group. That is why we do the sudo -s -u $USER ...
+      # Remove the pre-installed go version. We install the exact version we need.
+      run: |
+        sudo -s -u $USER --preserve-env bash ${{ github.workspace }}/hack/e2e/ci-script.sh
```
lentzi90: nit: I think this is only needed for BMO because we use libvirt and need to be in the libvirt group. For IPAM I guess we do not need it and so it should be possible to just run the script without the `sudo -s -u ...`

test/e2e/config/e2e_conf.yaml
```diff
@@ -0,0 +1,214 @@
+# E2E test scenario using local dev images and manifests built from the source tree for following providers:
+# - cluster-api
+# - bootstrap kubeadm
+# - control-plane kubeadm
+# - metal3
+
+images:
+# Use local dev images built source tree;
```
lentzi90: The comment says that we use local dev images, but then we actually use tags that exists in the registry (`latest`). I am not sure we actually use what we think we use then. It would be much safer to use a tag that doesn't exist, so we will get an obvious error.
Now it would be possible that we load the image but then the pod is configured to always pull, checks the registry and gets the image from there "silently".

lentzi90 (commented):
Awesome!
I haven't gone through it all yet, but leaving the comments I had so far.
Hopefully continuing tomorrow 🙂

test/e2e/basic_test.go
```diff
@@ -0,0 +1,325 @@
...
+		// Create namespace for the test if it doesn't exist
+		ns := &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: namespace,
+			},
+		}
+		_ = cl.Create(ctx, ns)
+
+		// Clean up any stale test resources from previous test runs
+		cleanupTestResources(ctx, cl)
```
lentzi90: Why not do this in AfterEach instead? I think it makes more sense to cleanup there instead of at the start of each test.

test/e2e/basic_test.go
```diff
@@ -0,0 +1,325 @@
...
+	})
+
+	It("Should allocate an IPAddress via Metal3 IPClaim", func() {
+		cl := bootstrapClusterProxy.GetClient()
+
+		By("Creating an IPPool")
+		ipPool := createIPPool(ctx, bootstrapClusterProxy, "test-ippool-m3claim")
```
lentzi90: I understand the desire to get rid of the big object with all the parameters, but hiding all the details in `createIPPool` also makes it harder to understand what the test does. Perhaps it would make sense to add a few more parameters to the function, or use a partial object so that it is more obvious what kind of pool is created.
This is not a blocker in anyway, just a food for thought.

test/e2e/basic_test.go
```diff
@@ -0,0 +1,325 @@
...
+var (
+	ctx                 = context.TODO()
+	specName            = "ipam-e2e"
+	namespace           string
+	clusterctlLogFolder string
+)
+
+var _ = Describe("Metal3 IPAM basic functionality", Label("ipam", "basic"), func() {
+	BeforeEach(func() {
+		namespace = testNamespace()
```
lentzi90, line 23: I am struggling to see how this namespace is used since it is not passed as parameter when creating IPPools or claims. Passing it around would be slightly more code, but it would make the functions much more usable in other situations. For example, what if we add a test at some point where we check some cross-namespace functionality? At that point, we would need to say which namespace the claim should be in. If the namespace is a parameter to the function, this is trivial to solve, but here we end up with a refactor of all functions that have this variable "hard-coded" in.
More a food for thought than requirement to change.

test/e2e/basic_test.go
```diff
@@ -0,0 +1,325 @@
...
+		cl := bootstrapClusterProxy.GetClient()
+
+		By("Creating an IPPool")
+		ipPool := createIPPool(ctx, bootstrapClusterProxy, "test-ippool-m3claim")
+
+		By("Verifying that the IPPool is created successfully")
+		verifyIPPool(ctx, bootstrapClusterProxy, ipPool)
+
+		By("Creating a Metal3 IPClaim referencing the pool")
+		claimName := fmt.Sprintf("test-ipclaim-%d", GinkgoParallelProcess())
```
lentzi90: Just checking that I understand. We are using separate namespaces already, right? So this is a sort of extra precaution or simply unnecessary?
This is probably just individual preference, but I would rather have the claim name be something related to the test itself and let the namespace handle the isolation.

test/e2e/basic_test.go
```diff
@@ -0,0 +1,325 @@
...
+
+	Expect(clusterProxy.GetClient().Create(ctx, claim)).To(Succeed())
+	Logf("Successfully created CAPI IPAddressClaim %s/%s", claim.Namespace, claim.Name)
+
+	return claim
+}
+
+// cleanupTestResources deletes any stale test resources from previous test runs
+// to prevent conflicts between parallel test executions.
+func cleanupTestResources(ctx context.Context, cl client.Client) {
```
lentzi90: Would it be possible to just delete the namespace? I can understand the need for checking that other resources can be cleaned up correctly first, but in the end it would be nice to just delete the namespace as a catch-all to get rid of anything we may have forgotten.

test/e2e/common.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func Logf(format string, a ...any) {
+	fmt.Fprintf(GinkgoWriter, "INFO: "+format+"\n", a...)
+}
+
+// testNamespace returns a namespace unique to the current parallel Ginkgo node,
+// preventing parallel test interference when running with multiple nodes.
+func testNamespace() string {
+	return fmt.Sprintf("ipam-e2e-%d", GinkgoParallelProcess())
```
lentzi90: Idea for improvement (maybe in a separate PR): Take a parameter with the spec name. That way it will be much easier to see what resources belong to what test.

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,239 @@
...
+	bootstrapClusterProxy framework.ClusterProxy
+
+	osType string
+
+	kubeconfigPath string
+	e2eTestsPath   string
+
+	numberOfControlplane int
+	numberOfWorkers      int
+	numberOfAllBmh       int
```
lentzi90: Do we need these?

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,239 @@
...
+	flag.BoolVar(&skipCleanup, "e2e.skip-resource-cleanup", false, "if true, the resource cleanup after tests will be skipped")
+	flag.BoolVar(&upgradeTest, "e2e.trigger-upgrade-test", false, "if true, the e2e upgrade test will be triggered and other tests will be skipped")
+	flag.BoolVar(&useExistingCluster, "e2e.use-existing-cluster", true, "if true, the test uses the current cluster instead of creating a new one (default discovery rules apply)")
+	flag.StringVar(&kubeconfigPath, "e2e.kubeconfig-path", os.Getenv("HOME")+"/.kube/config", "if e2e.use-existing-cluster is true, path to the kubeconfig file")
+	e2eTestsPath = getE2eTestsPath()
+
+	osType = strings.ToLower(os.Getenv("OS"))
+	if osType == "" {
+		osType = "ubuntu" // Default to linux if OS env var is not set
+	}
```
lentzi90: I think this is not needed

test/e2e/config/e2e_conf.yaml
lentzi90, line 1: I think there are quite many variables and intervals that are not used/needed in this config. We could probably also remove a lot of the providers. We do not test any upgrades for example, so only one version of each will ever be used.

lentzi90 (commented):
Some more comments below.
I think this is quite good already. The only thing I would want before approving is to clean up things we don't need. There are quite a lot of things that seem to be just copied from what we have in CAPM3/BMO. Let's get rid of them if we don't need them.

I also added many comments with ideas and suggestions. Feel free to ignore these, or maybe add them to some todo list.

Overall, this is really nice work! For next time, I would maybe suggest starting with something a bit smaller. For example just the scaffolding and one very basic test. That way it become less intimidating to review, and less time consuming to fix if there are many changes needed.

.github/workflows/e2e-test.yml
peppi-lotta: removed

test/e2e/config/e2e_conf.yaml
peppi-lotta: Removed the comment

test/e2e/basic_test.go
peppi-lotta: That does make more sense. Moving the `cleanupTestResources` into `AfterEach`

test/e2e/config/e2e_conf.yaml
peppi-lotta: Removed all unused variables etc.

test/e2e/e2e_suite_test.go
peppi-lotta: Removed all unused variables etc.

test/e2e/e2e_suite_test.go
peppi-lotta: Removed all unused variables etc.

peppi-lotta:
I addresses some of the comment made by @lentzi90. I will continue to look at the rest next week.

Makefile
```diff
@@ -377,9 +382,77 @@ verify-boilerplate:
...
+E2E_BMCS_CONF_FILE ?= $(ROOT_DIR)/test/e2e/config/bmcs-redfish-virtualmedia.yaml
+USE_EXISTING_CLUSTER ?= false
+SKIP_RESOURCE_CLEANUP ?= false
+GINKGO_NOCOLOR ?= false
+
+GOLANGCI_LINT_BIN := golangci-lint
+GOLANGCI_LINT_VER := v2.1.6
+GOLANGCI_LINT := $(abspath $(TOOLS_BIN_DIR)/$(GOLANGCI_LINT_BIN))
+GOLANGCI_LINT_PKG := github.com/golangci/golangci-lint/v2/cmd/golangci-lint
+
```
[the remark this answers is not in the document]
peppi-lotta: Removed. This was also clean up I missed when copying Things from CAPM3.

test/e2e/config/e2e_conf.yaml
```diff
@@ -0,0 +1,213 @@
...
+  - name: "v1.14.99"
+    value: "https://github.com/metal3-io/ip-address-manager/releases/download/v1.13.0/ipam-components.yaml"
+    type: "url"
+    contract: v1beta2
+    replacements:
+    - old: --metrics-addr=127.0.0.1:8080
+      new: --metrics-addr=:8080
+    files:
+    - sourcePath: "../data/shared/ipam-metal3/v1.14/metadata.yaml"
+      targetName: "metadata.yaml"
```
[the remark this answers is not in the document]
peppi-lotta: This should be pointing to local resources:

```
  - name: "v1.14.99"
    value: "../../../config/default"
    type: "kustomize"
    contract: v1beta2
    replacements:
    - old: --metrics-addr=127.0.0.1:8080
      new: --metrics-addr=:8080
    files:
    - sourcePath: "../../../metadata.yaml"
      targetName: "metadata.yaml"
```

Fixed.

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+	kubeconfigPath string
+)
+
+func init() {
+	flag.StringVar(&configPath, "e2e.config", "", "path to the e2e config file")
+	flag.StringVar(&artifactFolder, "e2e.artifacts-folder", "", "folder where e2e test artifact should be stored")
+	flag.BoolVar(&skipCleanup, "e2e.skip-resource-cleanup", false, "if true, the resource cleanup after tests will be skipped")
+	flag.BoolVar(&useExistingCluster, "e2e.use-existing-cluster", true, "if true, the test uses the current cluster instead of creating a new one (default discovery rules apply)")
+	flag.StringVar(&kubeconfigPath, "e2e.kubeconfig-path", os.Getenv("HOME")+"/.kube/config", "if e2e.use-existing-cluster is true, path to the kubeconfig file")
```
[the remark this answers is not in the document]
peppi-lotta: CAPM3 tests use these defaults. I think the CI test utilize this. I agree false is a better default here because we don't have any outside setting up before these e2e tests are run. It is nice to allow testing with an existing cluster though but let's leave it to the user to set the values.

test/e2e/basic_test.go
peppi-lotta: This works as well. I'm going to go with this approach as it is also cleaner and more concise code wise.

test/e2e/basic_test.go
peppi-lotta: The `namespace` is used in the `createIPPool` and `createIPClaim` functions. It is used so that there are no namespace collisions during the parallel test tuns. Based on this and other feed back I've better parametrized the usage of the `namespace` variable.

test/e2e/basic_test.go
peppi-lotta: IPPool needs to be created in every test case, so I feel it's nicer to have the function. I've now replaced the completely abstracted function with  a function that ingests a `CreateIPPoolInput` type. This way the values can be switched around more easily per test but I think the test remains clearer and easier to read this way.

test/e2e/basic_test.go
peppi-lotta: I think that approach would be nicer. I've implemented it.

test/e2e/common.go
peppi-lotta: I've changed the function so that forms a namespace based on prefix, spec name and, hash. Prefix is "e2e-", spec name is a cleand version of the `It` spec name and hash is a 6 letter unique ending to allow one test to have multiple namespaces if needed.

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,48 @@
...
+    e2e-tests:
+      type: string
+      default: "all"
+      description: "Which e2e test suite to run: basic, features, or all"
+    ref:
+      type: string
+      default: ${{ github.ref }}
+    timeout-minutes:
+      type: number
+      default: 90
```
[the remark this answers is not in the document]
peppi-lotta: fixed

hack/e2e/ci-script.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
+
+# -----------------------------------------------------------------------------
+# Description: This script sets up the environment and runs E2E tests for the
+#              IPAM project.
+# Usage:       From the root of the repo, run:
+#              ./test/e2e/ci-e2e.sh
```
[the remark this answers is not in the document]
peppi-lotta: fixed

test/e2e/feature_test.go
```diff
@@ -0,0 +1,661 @@
...
+	BeforeEach(func() {
+		namespace = testNamespace()
+		validateGlobals(specName)
+		cl := bootstrapClusterProxy.GetClient()
+		ns := &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: testNamespace(),
+			},
+		}
+		_ = cl.Create(ctx, ns)
```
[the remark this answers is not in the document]
peppi-lotta: fixed

test/e2e/basic_test.go
```diff
@@ -0,0 +1,326 @@
...
+	// Clean up stale IPClaims (Metal3)
+	ipClaimList := &ipamv1.IPClaimList{}
+	if err := cl.List(ctx, ipClaimList, client.InNamespace(namespace)); err == nil {
+		for i := range ipClaimList.Items {
```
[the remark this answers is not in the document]
peppi-lotta: I switched up the `cleanupTestResources` so that it removes the whole namespace of the test. This is no longer valid.

test/e2e/basic_test.go
```diff
@@ -0,0 +1,326 @@
...
+		namespace = testNamespace()
+		validateGlobals(specName)
+		cl := bootstrapClusterProxy.GetClient()
+		// Create namespace for the test if it doesn't exist
+		ns := &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: namespace,
+			},
+		}
+		_ = cl.Create(ctx, ns)
```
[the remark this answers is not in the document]
peppi-lotta: fixed

hack/e2e/ci-script.sh
```diff
@@ -0,0 +1,74 @@
...
+# -----------------------------------------------------------------------------
+
+set -eux
+
+REPO_ROOT=$(realpath "$(dirname "${BASH_SOURCE[0]}")/../..")
+cd "${REPO_ROOT}" || exit 1
```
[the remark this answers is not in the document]
peppi-lotta: Switched to use trap

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+// Using a SynchronizedBeforeSuite for controlling how to create resources shared across ParallelNodes (~ginkgo threads).
+// The local clusterctl repository & the bootstrap cluster are created once and shared across all the tests.
+var _ = SynchronizedBeforeSuite(func() []byte {
+	// Before all ParallelNodes.
+
+	Expect(configPath).To(BeAnExistingFile(), "Invalid test suite argument. e2e.config should be an existing file.")
+	Expect(os.RemoveAll(artifactFolder)).To(Succeed())
+	Expect(os.MkdirAll(artifactFolder, 0755)).To(Succeed(), "Invalid test suite argument. Can't create e2e.artifacts-folder %q", artifactFolder)
+
```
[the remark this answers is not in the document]
peppi-lotta: fixed

hack/e2e/ci-script.sh
```diff
@@ -0,0 +1,74 @@
...
+    exit 1
+    ;;
+esac
+
+# Run the e2e tests
+make test-e2e
+test_status="$?"
+
+# Collect all artifacts
+tar --directory "${REPO_ROOT}/test/e2e/_artifacts/" -czf "artifacts-e2e-ipam.tar.gz" "${REPO_ROOT}/test/e2e/_artifacts/"
```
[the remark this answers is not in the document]
peppi-lotta: fixed

docs/e2e.md
```diff
@@ -0,0 +1,124 @@
...
+| `features` | Advanced functionality: preallocations, pool exhaustion/recovery, multi-pool, status tracking, mixed claims |
+
+## Configuration
+
+### Make variables
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| `GINKGO_NODES` | `2` | Number of parallel Ginkgo nodes |
+| `GINKGO_TIMEOUT` | `3h` | Overall test timeout |
```
lentzi90: Considering the whole suite runs in less than 5 minutes on my laptop I think we can put something lower here 😂 
I would also consider not writing out all this here and just point to the Makefile instead. We will definitely forget to update the docs when making changes in the Makefile.

docs/e2e.md
```diff
@@ -0,0 +1,124 @@
...
+```text
+test/e2e/
+├── config/
+│   └── e2e_conf.yaml       # Provider and image configuration
+├── data/                    # Shared test data (metadata files)
+├── e2e_suite_test.go        # Suite setup: bootstrap cluster, install providers
+├── basic_test.go            # Basic label tests
+├── feature_test.go          # Features label tests
+└── common.go                # Shared helpers (logging, namespace, cleanup)
+```
```
lentzi90: I would skip this. It will quickly become outdated if we work actively on the tests, and the structure should hopefully be quite obvious from the filesystem directly.

docs/e2e.md
```diff
@@ -0,0 +1,124 @@
...
+1. Collects artifacts into a tarball
+
+Label filtering is controlled via the `GINKGO_FOCUS_LABELS` environment variable:
+
+```sh
+# Run basic tests only in CI
+GINKGO_FOCUS_LABELS=basic ./hack/e2e/ci-script.sh
+
+# Run feature tests only in CI
+GINKGO_FOCUS_LABELS=feature ./hack/e2e/ci-script.sh
```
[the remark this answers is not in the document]
peppi-lotta: Fixed

test/e2e/config/e2e_conf.yaml
```diff
@@ -0,0 +1,80 @@
+# E2E test scenario using local dev images and manifests built from the source tree for following providers:
+# - cluster-api
+# - bootstrap kubeadm
+# - control-plane kubeadm
+# - metal3
+
+images:
+- name: quay.io/metal3-io/ip-address-manager:latest
```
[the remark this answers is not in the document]
peppi-lotta: Fixed

test/e2e/basic_test.go
```diff
@@ -0,0 +1,181 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	capipamv1 "sigs.k8s.io/cluster-api/api/ipam/v1beta2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	ctx      = context.TODO()
+	specName = "ipam-e2e"
```
lentzi90: This global variable is used for all tests. As such, it doesn't really make sense. Either we should set a spec name for each test or we can just skip this all together. I would go with the latter for now. We only use it for getting "intervals", and they anyway fall back to `default`. So we could just delete this and change the intervals to use `default` directly.

lentzi90 (commented):
No blockers. I'd rather take this in than keep reviewing the 2000+ lines for every little change.
A few comments below. I also want to mention I tested this on my laptop and it worked great! There was one issue with the kustomize version and some patches that are pre-existing. I'll raise an issue for that separately.
I also noticed that log collection worked, however, we do not gather any yaml. This is something we should add. It will be needed for debugging when tests start failing.
/approve

test/e2e/common.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// cleanupNamespace deletes the test namespace, which cascade-deletes all resources within it.
+func cleanupNamespace(ctx context.Context, cl client.Client, namespace string) {
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespace,
+		},
+	}
+	if err := cl.Delete(ctx, ns); err != nil && !apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
peppi-lotta: Skipping all namespace clean up when this is enabled.

docs/e2e.md
peppi-lotta: Maybe I can lower it a little bit... :zany_face:

docs/e2e.md
peppi-lotta: Removed

test/e2e/basic_test.go
peppi-lotta: Removed `specName`

peppi-lotta:
@lentzi90 I added e2e PR test. Any idea why it get queued but never starts?

config/default/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: quay.io/metal3-io/ip-address-manager:main
+      - image: "quay.io/metal3-io/ip-address-manager:e2e-test"
```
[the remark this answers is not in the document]
lentzi90: This is true. It would be better to leave the default as is and create a separate overlay for e2e.

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,47 @@
...
+      runner:
+        type: string
+        default: "oracle-vm-4cpu-16gb-x86-64"
+      ginkgo-focus-labels:
+        type: string
+        default: ""
+        description: "Ginkgo label filter expression (e.g. 'basic', 'features'). Empty runs all tests."
+      ref:
+        type: string
+        default: ${{ github.ref }}
```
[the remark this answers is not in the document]
peppi-lotta: This is following convention from BMO project. I'm going to keep it as is.

config/default/manager_image_patch.yaml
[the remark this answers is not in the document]
peppi-lotta: I added this `MANIFEST_IMG=$(CONTROLLER_IMG) MANIFEST_TAG=e2e-test $(MAKE) set-manifest-image` in the e2e process to get local builds working properly as well. CAPM3 does these rewrites as well, that you just have to remember not to push if you run tests locally. I was thinking it would work similarly here and then accidentally pushed the change when I was not supposed to, showcasing why this is not good approach. I added the overlay now.

test/e2e/common.go
```diff
@@ -0,0 +1,187 @@
...
+	nonAlphaNum = regexp.MustCompile(`[^a-z0-9]+`)
+)
+
+func Logf(format string, a ...any) {
+	fmt.Fprintf(GinkgoWriter, "INFO: "+format+"\n", a...)
+}
+
+// testNamespace generates a valid Kubernetes namespace name based on the current It spec name.
+// Uses a prefix + sanitized spec text + hash suffix to stay within the 63-char limit.
+func testNamespace() string {
```
adilGhaffarDev, line 33: I am not sure if we need to have this complicated logic just for namespace.
Can you explain why this is needed?

lentzi90:
> @lentzi90 I added e2e PR test. Any idea why it get queued but never starts?

Sorry for the late reply :sweat_smile: 
I guess this was solved, since I see the test passed

test/e2e/common.go
peppi-lotta: Each test has it's own namespace so clean up can be done by deleting a namespace only. Tests are run in parallel so if they have colliding namespaces there are errors when wrong parts get cleaned up at the wrong time. Also there is an option to run without clean up and having all resources in a namespace which name is based on the test name made the debugging  seem easier. The `testNamespace()` function adds the hash just in case if we need multiple namespaces for one test in any future testcase. 

We could write separate namespace for each test but I just felt like this function approach was quite nice and easily reusable.

hack/e2e/ensure_go.sh
```diff
@@ -0,0 +1,51 @@
...
+MINIMUM_GO_VERSION=go1.25.10
+
+# Ensure the go tool exists and is a viable version, or installs it
+verify_go_version()
+{
+    # If go is not available on the path, get it
+    if ! [ -x "$(command -v go)" ]; then
+        if [[ "${OSTYPE}" == "linux-gnu" ]]; then
+            echo 'go not found, installing'
+            curl -sLo "/tmp/${MINIMUM_GO_VERSION}.linux-amd64.tar.gz" "https://go.dev/dl/${MINIMUM_GO_VERSION}.linux-amd64.tar.gz"
```
kashifest, line 30: Just a note for future, we should perhaps make these scripts also to work on arm64 archs if this is also meant to be run/used locally.

kashifest (commented):
Awesome job @peppi-lotta 
/lgtm

PR #1436, opened 2026-06-08
🌱 Bump x/net to v0.55.0

UncivilizedGhost:
**What this PR does / why we need it**:Xnet needs to be manually bumped as [dependabot does not handle indirect dependencies](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated#allowing-specific-dependencies-to-be-updated).

Fixes #

**Checklist:**


test/go.mod
```diff
@@ -0,0 +1,76 @@
+module github.com/metal3-io/ip-address-manager/test
```
peppi-lotta: Remove this file

test/go.sum
```diff
@@ -0,0 +1,176 @@
+github.com/Masterminds/semver/v3 v3.4.0 h1:Zog+i5UMtVoCU8oKka5P7i9q9HgrJeGzI9SA1Xbatp0=
```
peppi-lotta: Remove this file

test/go.mod
kashifest: Why is this new file being added here as part of this commit?

peppi-lotta (commented):
The `test/go.mod` and `test/go.sum` are new files that are only in main. They should not be added for branch-1.13. 

The files were added to main because of the new e2e tests added.

test/go.mod
peppi-lotta: Apparently we were reviewing at the same time. Check my comment: https://github.com/metal3-io/ip-address-manager/pull/1436#pullrequestreview-4448319698

peppi-lotta:
Now this looks good!

/approve
/test metal3-centos-e2e-integration-test-release-1-13

kashifest (commented):
/approve

kashifest (commented):
/lgtm

PR #1439, opened 2026-06-08
:bug: Add uniqueness check for Preallocation at IPPool creation and validation times

### LINKED ISSUE #1440: IPPool accepts preAllocations where same IP is allocated to multiple claims

### Step 1 — Create an IPPool with duplicate preAllocations

```bash
kind create cluster
clusterctl init --infrastructure docker --ipam metal3
kubectl create namespace test-ns

kubectl apply -f - <<EOF
kind: IPPool
metadata:
  name: test-pool
  namespace: test-ns
spec:
  namePrefix: "test-ip"
  pools:
    - start: "192.168.1.10"
      end: "192.168.1.50"
      prefix: 24
      gateway: "192.168.1.1"
  preAllocations:
    claim-1: "192.168.1.10"
    claim-2: "192.168.1.10"    # <-- Duplicate value (same IP, different key)
    claim-3: "192.168.1.11"
  prefix: 24
  gateway: "192.168.1.1"
EOF
```

The IPPool is created without validation errors, even though
`192.168.1.10` is listed twice.

### Step 2 — Create the first IPClaim for the duplicated IP

```bash
kubectl apply -f - <<EOF
kind: IPClaim
metadata:
  name: claim-1
  namespace: test-ns
spec:
  pool:
    name: test-pool
    namespace: test-ns
EOF
```

### Step 3 — Check which IP was allocated to claim-1

```bash
kubectl describe ipclaim claim-1 -n test-ns
kubectl get ipaddress.ipam.metal3.io -n test-ns
```

`claim-1` gets `192.168.1.10` as expected.

### Step 4 — Create a second IPClaim targeting the same duplicated IP

```bash
kubectl apply -f - <<EOF
kind: IPClaim
metadata:
  name: claim-2
  namespace: test-ns
spec:
  pool:
    name: test-pool
    namespace: test-ns
EOF
```

### Step 5 — Observe that claim-2 silently fails

```bash
kubectl describe ipclaim claim-2 -n test-ns
kubectl get ipaddress.ipam.metal3.io -n test-ns
```

`claim-2` has no IP allocated and no error condition — it is stuck.

### Step 6 — Check controller logs

```bash
kubectl logs -n metal3-ipam-system deployment/ipam-controller-manager \
  | grep -i "claim-2\|collision\|error"
```

No meaningful error message is produced for the silent failure.

## Expected behavior

The `kubectl apply` in Step 1 is rejected by the validating webhook
with an error indicating the duplicate IP:

```
The IPPool "test-pool" is invalid:
  spec.preAllocations.claim-2: Invalid value: "192.168.1.10":
    IP address is already pre-allocated to claim "claim-1"
```

When creating an IPPool with duplicate IP addresses in `preAllocations` (same IP mapped to different claim names), the webhook accepts the resource without error. At runtime the second claim silently fails to allocate.

There is no valid use-case for pre-allocating the same IP to multiple claims.

Add validation in `validatePoolRanges()` (called by both `ValidateCreate` and `ValidateUpdate`) that rejects IPPools where multiple preAllocation entries map to the same IP address.

**Environment:**

- Cluster-api version: 1.13
- IPAM version: 1.13
- environment: Kind cluster with clusterctl setup

/kind bug

peppi-lotta:
# Add uniqueness validation for IPPool preAllocation IP addresses

## Problem

When creating an IPPool with duplicate IP addresses in `preAllocations` (same IP mapped to different claim names), the webhook accepts the resource without error. At runtime the second claim silently fails to allocate.

There is no valid use-case for pre-allocating the same IP to multiple claims.

## Fix

Add validation in `validatePoolRanges()` (called by both `ValidateCreate` and `ValidateUpdate`) that rejects IPPools where multiple preAllocation entries map to the same IP address.

## Expected behavior after this PR

The `kubectl apply` in Step 1 is rejected by the validating webhook
with an error indicating the duplicate IP:

```
The IPPool "test-pool" is invalid:
  spec.preAllocations.claim-2: Invalid value: "192.168.1.10":
    IP address is already pre-allocated to claim "claim-1"
```

fixes #1440

**Checklist:**

### DIFF
--- a/internal/webhooks/v1alpha1/ippool_webhook.go
+++ b/internal/webhooks/v1alpha1/ippool_webhook.go
@@ -249,7 +249,8 @@ func (webhook *IPPool) validatePoolRanges(pool *ipamv1.IPPool) field.ErrorList {
 		}
 	}
 
-	// Validate preAllocations IPs
+	// Validate preAllocations IPs.
+	ipToClaimName := make(map[netip.Addr]string, len(pool.Spec.PreAllocations))
 	for name, ipAddr := range pool.Spec.PreAllocations {
 		if err := validateIPAddress(ipAddr); err != nil {
 			allErrs = append(allErrs,
@@ -259,6 +260,23 @@ func (webhook *IPPool) validatePoolRanges(pool *ipamv1.IPPool) field.ErrorList {
 					"is not a valid IP address",
 				),
 			)
+			continue
+		}
+
+		// Use netip.Addr.String() form as map key so that equivalent
+		// addresses with different textual representations (e.g. IPv6 with/without
+		// leading zeros) are detected as duplicates.
+		ipAddress, _ := netip.ParseAddr(string(ipAddr))
+		if existingClaim, exists := ipToClaimName[ipAddress]; exists {
+			allErrs = append(allErrs,
+				field.Invalid(
+					field.NewPath("spec", "preAllocations", name),
+					ipAddr,
+					fmt.Sprintf("IP address is already pre-allocated to claim %q", existingClaim),
+				),
+			)
+		} else {
+			ipToClaimName[ipAddress] = name
 		}
 	}
 
--- a/internal/webhooks/v1alpha1/ippool_webhook_test.go
+++ b/internal/webhooks/v1alpha1/ippool_webhook_test.go
@@ -308,6 +308,22 @@ func TestIPPoolValidation(t *testing.T) {
 			},
 		},
 		{
+			name:      "should succeed with unique preAllocation IPs",
+			expectErr: false,
+			c: &ipamv1.IPPool{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "foo",
+				},
+				Spec: ipamv1.IPPoolSpec{
+					PreAllocations: map[string]ipamv1.IPAddressStr{
+						"claim1": "192.168.0.10",
+						"claim2": "192.168.0.11",
+						"claim3": "192.168.0.12",
+					},
+				},
+			},
+		},
+		{
 			name:      "should succeed with random strategy when pool is bounded by start and end",
 			expectErr: false,
 			c: &ipamv1.IPPool{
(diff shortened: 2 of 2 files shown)


internal/webhooks/v1alpha1/ippool_webhook.go
```diff
@@ -225,6 +225,7 @@ func (webhook *IPPool) validatePoolRanges(pool *ipamv1.IPPool) field.ErrorList {
 	}
 
 	// Validate preAllocations IPs
+	ipToClaimName := make(map[ipamv1.IPAddressStr]string, len(pool.Spec.PreAllocations))
 	for name, ipAddr := range pool.Spec.PreAllocations {
 		if err := validateIP(ipAddr); err != nil {
 			allErrs = append(allErrs,
```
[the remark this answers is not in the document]
peppi-lotta: Added check and unit tests

lentzi90 (commented):
/approve

internal/webhooks/v1alpha1/ippool_webhook.go
```diff
@@ -260,6 +264,18 @@ func (webhook *IPPool) validatePoolRanges(pool *ipamv1.IPPool) field.ErrorList {
 				),
 			)
 		}
+		ipAddress, _ := netip.ParseAddr(string(ipAddr))
```
tuminoid, line 269: If there are invalid ips, we don't check error here, the ipAddress is empty, and its on L277 used as index, so two invalid ips both fail (no harm) but the error becomes the invalid "" ip is double-preallocated.

tuminoid (commented):
Minor comment.

internal/webhooks/v1alpha1/ippool_webhook.go
kashifest: isn't that checked already in line L258 in `validateIPAddress(ipAddr)` ?

tuminoid: That doesn't exit or continue, so it reaches this and following lines too. We accumulate ton of errors, including the false ones but never return.

kashifest: Ah right! so we just validate IP address and for invalid ones we append the error thats it and then we continue, thats bad, good catch

peppi-lotta: This is a good finding! I'll add a `continue` in the `validateIPAddress(ipAddr)` so that duplicate check is skipped if an IP has more severe issues.

tuminoid (commented):
/lgtm

PR #1446, opened 2026-06-10
🌱 Removed zizmor inline ignore

migiFi:
Removed the `# zizmor: ignore[artipacked]` inline suppression comment from `.github/workflows/release.yaml`. The `artipacked` warning is now handled via a targeted ignore rule in `.zizmor.yml` at the repo root instead, since `persist-credentials: false` cannot be used here as credentials are needed later in the workflow.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -20,7 +20,7 @@ jobs:
     if: github.repository == 'metal3-io/ip-address-manager'
     steps:
     - name: Checkout code
-      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2 # zizmor: ignore[artipacked]
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -20,9 +20,10 @@ jobs:
     if: github.repository == 'metal3-io/ip-address-manager'
     steps:
     - name: Checkout code
-      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2 # zizmor: ignore[artipacked]
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       with:
         fetch-depth: 0
+        persist-credentials: false
```
tuminoid: Nope, the reason the artipacked is there that we need to credentials later in the flow.

tuminoid (commented):
This is wrong fix. The fix is moving the comment (if allowed) to previous line, or creating a targeted zizmor rule.

/hold

tuminoid (commented):
/approve

PR #1448, opened 2026-06-12
🌱 Removed zizmor inline ignore & set dependabot cooldown to 3

migiFi:
Removed the `# zizmor: ignore[artipacked]` inline suppression comment from `.github/workflows/release.yaml`. In addition, the cooldown on dependabot was changed from 7 days to 3 days. Instead of suppressing the zizmor warning with `zizmor: ignored`, this adds a proper `.zizmor.yml` config at the repo root.  Both of the ignore/warnings are now handled via its own targeted ignore rule in `.zizmor.yml` at the repo root instead. The workflow at `.github/workflows/zizmor.yml` picks it up via the `config:` input.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,13 +20,13 @@ updates:
   - dependency-name: "*"
     update-types: ["version-update:semver-major"]
   cooldown:
-    default-days: 7
+    default-days: 3
   commit-message:
     prefix: ":seedling:"
   labels:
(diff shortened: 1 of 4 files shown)


tuminoid (commented):
/approve

kashifest (commented):
This PR needs rebase @migiFi

lentzi90 (commented):
/lgtm

PR #1471, opened 2026-06-23
🌱 Add image build workflow for PR

UncivilizedGhost:
**What this PR does / why we need it**: Utilize container-build-workflow from project-infra.

Fixes #

**Checklist:**

### DIFF
--- a/.github/workflows/build-images-pr.yml
+++ b/.github/workflows/build-images-pr.yml
@@ -1,0 +1,21 @@
+name: Build IPAM Image on PRs
+
+on:
+  pull_request:
+    types: [opened, edited, reopened, synchronize, ready_for_review]
+
+permissions:
+  contents: read
+
+jobs:
+  build_ipam:
+    name: Build IPAM container image
+    if: github.repository == 'metal3-io/ip-address-manager'
+    permissions:
+      contents: read
+    uses: metal3-io/project-infra/.github/workflows/container-image-build.yml@main # zizmor: ignore[unpinned-uses]
+    with:
+      image-name: 'ip-address-manager'
+      pushImage: false
+      generate-sbom: false
+      sign-image: false


lentzi90 (commented):
Do we need the multi-arch matrix? I think that is only needed for Ironic. IPAM and other controllers can be cross-compiled easily. I have not looked deeply into the image building pipeline though so I could be mistaken.

UncivilizedGhost:
From what I can tell it doesn't actually need it. I've amended the simplified version.

.github/workflows/build-images-pr.yml
```diff
@@ -0,0 +1,31 @@
...
+    types: [opened, edited, reopened, synchronize, ready_for_review]
+
+permissions: {}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
```
tuminoid: Could we use the container-build-workflow from project-infra for consistency?

UncivilizedGhost: I have been basing this on the workflow in [ironic image](https://github.com/metal3-io/ironic-image/blob/main/.github/workflows/ironic-image-build-pr.yml). The ticket I'm tackling aims to add similar workflows to several repositories. This approach has the small benefit of caching but I don't know if they are any other reasons to switch to the container worflow

Rozzii: Lets keep consistency, for the merge trigger we use this https://github.com/metal3-io/ip-address-manager/blob/main/.github/workflows/build-images-action.yml, please base this PR test on the same logic, skip pushing and signing though.

Rozzii (commented):
Lets keep consistency, for the merge trigger we use this https://github.com/metal3-io/ip-address-manager/blob/main/.github/workflows/build-images-action.yml, please base this PR test on the same logic, skip pushing and signing though.

.github/workflows/build-images-pr.yml
```diff
@@ -0,0 +1,21 @@
...
+permissions:
+  contents: read
+
+jobs:
+  build_ipam:
+    name: Build IPAM container image
+    if: github.repository == 'metal3-io/ip-address-manager'
+    permissions:
+      contents: read
+    uses: metal3-io/project-infra/.github/workflows/container-image-build.yml@main # zizmor: ignore[unpinned-uses]
```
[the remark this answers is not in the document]
UncivilizedGhost: We are intentionally tracking main to ensure we always get the latest updates and fixes

lentzi90 (commented):
/lgtm

tuminoid (commented):
/approve

PR #1488, opened 2026-07-01
:seedling:  Pin gcrane in verify-release.sh and add renovate to update it

peppi-lotta:
Pin gcrane with image digest in `verify-release.sh` and add renovate to update the pinned version. The reason gcrane image is not pinned is that the images don't get maintained too long and it has been easier to leave it unpinned. I'm suggesting we pin the version and add renovate to upgrade the version when there is a new on. I could get a clear picture how long the images are actually available but updating to new version should be at least often enough to not have the digest go stale on us. 

Fixes #

**Checklist:**

### DIFF
--- a/hack/renovate-validator.sh
+++ b/hack/renovate-validator.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+# Validates renovate.json configuration using renovate-config-validator.
+# Requires Node.js 22+ for Renovate v40.
+
+set -eux
+
+IS_CONTAINER="${IS_CONTAINER:-false}"
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
+WORKDIR="${WORKDIR:-/workdir}"
+
+if [ "${IS_CONTAINER}" != "false" ]; then
+    npx --yes -p renovate renovate-config-validator
+else
+    "${CONTAINER_RUNTIME}" run --rm \
+        --env IS_CONTAINER=TRUE \
+        --volume "${PWD}:${WORKDIR}:ro,z" \
+        --entrypoint sh \
+        --workdir "${WORKDIR}" \
+        docker.io/node:24-alpine \
+        "${WORKDIR}"/hack/renovate-validator.sh "$@"
+fi
--- a/hack/renovate-validator.sh
+++ b/hack/renovate-validator.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+# Validates renovate.json configuration using renovate-config-validator.
+# Requires Node.js 22+ for Renovate v40.
+
+set -eux
+
+IS_CONTAINER="${IS_CONTAINER:-false}"
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
+
+if [ "${IS_CONTAINER}" != "false" ]; then
+    npx --yes -p renovate renovate-config-validator
+else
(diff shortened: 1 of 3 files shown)


hack/renovate-validator.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+# Validates renovate.json configuration using renovate-config-validator.
+# Requires Node.js 22+ for Renovate v40.
+
+set -eux
+
+IS_CONTAINER="${IS_CONTAINER:-false}"
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
+WORKDIR="${WORKDIR:-/workdir}"
```
tuminoid: We're removing this variable in other repos based on the comments in IRSO etc for hardening, so maybe lets just hardcoded as `/workdir`.

hack/verify-release.sh
```diff
@@ -138,7 +138,7 @@ if [[ -n "${CONTAINER_RUNTIME}" ]]; then
     declare -a GCRANE_CMD=(
         "${CONTAINER_RUNTIME}" run --rm
         --pull always
-        gcr.io/go-containerregistry/gcrane:latest
+        gcr.io/go-containerregistry/gcrane:v0.21.7@sha256:9f84ccb966ace4576dcd72543cc980056d037e499d5f389cd39a78e4b04740a6
```
tuminoid: Might be easier to write the renovate regex if the version+sha are in separate variable.

hack/renovate-validator.sh
peppi-lotta: Changed to hardcoded value

hack/verify-release.sh
peppi-lotta: I separated the SHA and version into environment variables and updated the regex.

kashifest (commented):
/approve

hack/renovate-validator.sh
```diff
@@ -0,0 +1,20 @@
...
+
+if [ "${IS_CONTAINER}" != "false" ]; then
+    npx --yes -p renovate renovate-config-validator
+else
+    "${CONTAINER_RUNTIME}" run --rm \
+        --env IS_CONTAINER=TRUE \
+        --volume "${PWD}:/workdir:ro,z" \
+        --entrypoint sh \
+        --workdir "/workdir" \
+        docker.io/node:24-alpine \
```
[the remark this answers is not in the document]
peppi-lotta: We don't pin `docker.io/golang:1.26` in `hack/gomod.sh`. We have decided this is okay for testing. So I don't think this needs to be pinned ether.

hack/renovate-validator.sh
```diff
@@ -0,0 +1,20 @@
...
+# Validates renovate.json configuration using renovate-config-validator.
+# Requires Node.js 22+ for Renovate v40.
+
+set -eux
+
+IS_CONTAINER="${IS_CONTAINER:-false}"
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
+
+if [ "${IS_CONTAINER}" != "false" ]; then
+    npx --yes -p renovate renovate-config-validator
```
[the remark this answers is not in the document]
peppi-lotta: By same logic we leave container image unpinned I think we can leave this version unpinned.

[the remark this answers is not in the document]
peppi-lotta: This is valid. I'll add this.

PR #1506, opened 2026-07-23
🐛 Fix off-by-one in deleteOwnerRefFromList

UncivilizedGhost:
**What this PR does / why we need it**:RefListLen was correct so slicing with (refListLen-1) dropped two elements instead of one, which discarded an additional element.

Fixes #

**Checklist:**

### DIFF
--- a/ipam/utils.go
+++ b/ipam/utils.go
@@ -120,9 +120,9 @@ func deleteOwnerRefFromList(refList []metav1.OwnerReference,
 	if len(refList) == 1 {
 		return []metav1.OwnerReference{}, nil
 	}
-	refListLen := len(refList) - 1
-	refList[index] = refList[refListLen]
-	refList, err = deleteOwnerRefFromList(refList[:refListLen-1], objType, objMeta)
+	refListLastIndex := len(refList) - 1
+	refList[index] = refList[refListLastIndex]
+	refList, err = deleteOwnerRefFromList(refList[:refListLastIndex], objType, objMeta)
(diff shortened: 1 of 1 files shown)


ipam/utils.go
```diff
@@ -122,7 +122,7 @@ func deleteOwnerRefFromList(refList []metav1.OwnerReference,
 	}
 	refListLen := len(refList) - 1
 	refList[index] = refList[refListLen]
-	refList, err = deleteOwnerRefFromList(refList[:refListLen-1], objType, objMeta)
```
peppi-lotta, line 125: I think based on variable naming more logical would be

```
refListLen := len(refList)
refList[index] = refList[refListLen-1]
refList, err = deleteOwnerRefFromList(refList[:refListLen-1], objType, objMeta)
```

peppi-lotta: Or

```
refListLastIndex := len(refList) - 1
refList[index] = refList[refListLastIndex]
refList, err = deleteOwnerRefFromList(refList[:refListLastIndex], objType, objMeta)
```

peppi-lotta (commented):
Just a nit. This does fix the one off issue and  I'm happy take this in as is. I added two suggestions I think might be more logical based on variable naming.

ipam/utils.go
UncivilizedGhost: Updated variable name to refListLastIndex

smoshiur1237 (commented):
/lgtm

PR #1526, opened 2026-07-29
🐛 Fix IPv4 overflow detection

UncivilizedGhost:
**What this PR does / why we need it**: The overflow check used byte-length and
fixed index checks (IPBytes[4], IPBytes[5]). This was unreliable

Replace with a direct comparison against the maximum IPv4 address.
Replaced magic number with a constant to represent byte length for linter

Fixes #

**Checklist:**

### DIFF
--- a/api/v1alpha1/utils.go
+++ b/api/v1alpha1/utils.go
@@ -24,6 +24,15 @@ import (
 	"net"
 )
 
+const (
+	// ipv6ByteLen is the byte length of an IPv6 address.
+	ipv6ByteLen = 16
+	// Big Int value of 255.255.255.255
+	// Maximum numeric value in hex format of the IPv4-Mapped IPv6 Address
+	// according to https://www.rfc-editor.org/rfc/rfc4291.html#section-2.5.5.2
+	maxIPv4Int = 0xFFFFFFFFFFFF
+)
+
 // GetIPAddress renders the IP address, taking the index, offset and step into
 // account, it is IP version agnostic.
 func GetIPAddress(entry Pool, index int) (IPAddressStr, error) {
@@ -182,11 +191,14 @@ func lastIPInSubnet(n *net.IPNet) net.IP {
 // Note that if the resulting IP address is in the format ::ffff:xxxx:xxxx then
 // ip.String will fail to select the correct type of ip.
 func addOffsetToIP(ip, endIP net.IP, offset int) (net.IP, error) {
-	ip4 := false
-	if ip.To4() != nil {
-		ip4 = true
+	// Detect IPv4 before normalizing to 16-byte form.
+	ip4 := ip.To4() != nil
+	// Normalize to the 16-byte form so the byte-length and value checks
+	// below hold for both ParseIP and ParseCIDR inputs.
+	ip = ip.To16()
+	if ip == nil {
+		return nil, errors.New("invalid IP address")
 	}
-
 	// Create big integers.
 	IPInt := big.NewInt(0)
 	OffsetInt := big.NewInt(int64(offset))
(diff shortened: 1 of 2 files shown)


api/v1alpha1/utils.go
```diff
@@ -202,9 +207,13 @@ func addOffsetToIP(ip, endIP net.IP, offset int) (net.IP, error) {
...
-	// Verify that the IPv4 or IPv6 fulfills theirs constraints.
-	if (ip4 && IPBytesLen > 6 && IPBytes[4] != 255 && IPBytes[5] != 255) ||
-		(!ip4 && IPBytesLen > 16) {
+	// Verify that the IPv4 or IPv6 fulfills their constraints.
+	if ip4 {
+		maxIPv4 := big.NewInt(0).SetBytes(net.ParseIP("255.255.255.255").To16())
+		if IPInt.Cmp(maxIPv4) > 0 {
+			return nil, fmt.Errorf("IP address overflow for : %s", ip.String())
+		}
+	} else if IPBytesLen > ipv6ByteLen {
```
Rozzii, line 222:
```suggestion
	if ip4 {
		if IPInt.Cmp(big.NewInt(maxIPv4Int)) > 0 {
			return nil, fmt.Errorf("IP address overflow for : %s", ip.String())
		}
	} else if IPBytesLen > ipv6ByteLen {
		return nil, fmt.Errorf("IP address overflow for : %s", ip.String())
	}
```

api/v1alpha1/utils.go
```diff
@@ -24,6 +24,11 @@ import (
 	"net"
 )
 
+const (
+	// ipv6ByteLen is the byte length of an IPv6 address.
+	ipv6ByteLen = 16
```
Rozzii, line 29:
```suggestion
	ipv6ByteLen = 16
	// Big Int value of 255.255.255.255
	maxIPv4Int = 0xFFFFFFFFFFFF
```

api/v1alpha1/utils.go
```diff
@@ -202,9 +207,13 @@ func addOffsetToIP(ip, endIP net.IP, offset int) (net.IP, error) {
```
Rozzii, line 214: On the top
ip4 := ip.To4() != nil
no need for that 4 lines

below that immediately:
// Normalize to the 16-byte form so the byte-length and value checks
// below hold for both ParseIP and ParseCIDR inputs.
ip = ip.To16()
	if ip == nil {
		return nil, errors.New("invalid IP address")
	}

Rozzii (commented):
Good effort, but we can simplify a bit.

api/v1alpha1/utils.go
Rozzii: you can also comment

// Maximum numeric value in hex format of the  IPv4-Mapped IPv6 Address according to https://www.rfc-editor.org/rfc/rfc4291.html#section-2.5.5.2

api/v1alpha1/utils.go
UncivilizedGhost: Isn't the  `ip4 := ip.To4() != nil` lines needed to ensure the comparison `Cmp(big.NewInt(maxIPv4Int)` only occurs for IP4 addresses since otherwise all ip6 addresses will be invalid ?

Rozzii: I mean have ip4 := ip.To4() != nil one liner not the 4 line format that we have now. But I get the confusion, my phrasing wasn't clear .

Rozzii (commented):
Write a unit test for this please, or extend the existing unit tests to test for minimum, maximum, overflow scenarios, so bounds, overflow, and 1 or 2 correct addresses both ipv4 and 6

Rozzii (commented):
Looks good!
/approve
/cc @peppi-lotta

smoshiur1237 (commented):
/lgtm

PR #1532, opened 2026-07-29
🌱 Remove --network=host from docker build targets

migiFi:
Remove the `--network=host` flag from docker-build, docker-build-e2e, and docker-build-debug targets in the Makefile. Docker's default network isolation is sufficient for downloading Go modules from proxy.golang.org.

This hardens the build environment by reducing unnecessary host network access during builds.

No functional changes, builds works without the flag.


lentzi90 (commented):
/approve

adilGhaffarDev (commented):
/lgtm

PR #1533, opened 2026-07-30
🌱 Harden hack scripts agaist WORKDIR volume mount injection

migiFi:
Affects the following scripts under /hack dir:
- codegen.sh
- gomod.sh
- unit.sh
- manifestlint.sh
- markdownlint.sh
- shellcheck.sh

This scripts read WORKDIR from then env and insert it directly into the conatiner's --volume mount specification. Becausee the volume spec is colon-bounded (`host:target:options`), a WORKDIR value cotaining a colon could alter how the mount is parsed.

This change removes the WORKDIR var and hardcodes the container path to /workdir, which matches the pattern already use in serveral other metal3 repos. Since /workdir was already the default there is no change in behavior for normal use, but the injection surface is eliminated.


Rozzii (commented):
/lgtm

Rozzii:

PR #1535, opened 2026-07-31
:seedling: Update image delimiter in kustomize patch commands to use '#' instead of '@'

peppi-lotta:
This should fix release process. Sed command now fails because delimiter is @ and image digest syntax contains that marker. Replace "@" with "#". I think we can also remove the `\"'"` safely. The environment variables are in single quotes and surrounded with `$()`.

Fixes #

**Checklist:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -280,13 +280,13 @@ docker-build-%:
 .PHONY: set-manifest-image
 set-manifest-image:
 	$(info Updating kustomize image patch file for manager resource)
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/lgtm

Rozzii (commented):
/approve

Rozzii:
/override  metal3-ubuntu-e2e-integration-test-main metal3-centos-e2e-integration-test-main

PR #1539, opened 2026-08-07
 🌱 Add v1.15 metadata

smoshiur1237:
Adding v1.15 metadata

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -2,6 +2,9 @@ apiVersion: clusterctl.cluster.x-k8s.io/v1alpha3
 kind: Metadata
 releaseSeries:
 - major: 1
+  minor: 15
+  contract: v1beta2
+- major: 1
   minor: 14
   contract: v1beta2
 - major: 1


smoshiur1237:
/hold
This will go in after new release v1.14 branch

Rozzii (commented):
Something went sideways with the CI, please check.

lentzi90:
I am not completely sure, but I think the e2e test config may need updating to include the new minor.
/cc @peppi-lotta

peppi-lotta:
> I am not completely sure, but I think the e2e test config may need updating to include the new minor. /cc @peppi-lotta

Yup! CAPI e2e framework is very strict and fails for versions that it can't find in upstream git. I think this error happens because they don't have the 1.15 in their [metadata.yaml](https://github.com/kubernetes-sigs/cluster-api/blob/main/metadata.yaml) yet...

Also I think this update can and should be done at the same time as updating the e2e config version (after 1.14-rc when 1.14 has it's own branch)

smoshiur1237:
Closing this as the metadata changes has landed in #1550

PR #1540, opened 2026-08-07
✨ Add capacity and availableIPCount to IPPool status

### LINKED ISSUE #161: Add total IP count and available IP count in IPPool.Status
**User Story**

As a developer/user I would like to get the total IP count and available IP count of an IPPool CR by checking its status field.

**Detailed Description**

[A clear and concise description of what you want to happen.]

[Miscellaneous information that will assist in solving the issue.]

/kind feature

migiFi:
Add `capacity` and `availableIPCount` fields to `IPPoolStatus` so pool capacity and remaining addresses are readable directly from the resource.

The controller computes the total as the sum of all pool sizes and the available count as the total minus the addresses currently in use, updating them on every reconcile. Both are also exposed as `kubectl get ippool` columns.

- Add `capacity`, `AvailableIPCount`, and `CapacityOverflow` to `IPPoolStatus` (with printer columns)
- Compute counts in `updateAddressCounts()`, wired into `UpdateAddresses()`
- Reuse the existing `GetPoolSize()` to size each pool (now returns a recognizable `ErrPoolSizeOverflow`)
- best-effort: if a pool's size cannot be determined at all, reconcile is not failed and the counts are left unset
- if the combined pool size is too large to represent as a int64 (e.g. a very large IPv6 subnet such as a /64 or larger), `Capacity` is capped at the int64 maximum, `AvailableIPCount` is left unset, and `CapacityOverflow` is set to true 

Previously, an IPPool's capacity and remaining addresses could only be derived by manually summing the spec ranges and subtracting the allocations. This PR adds `capacity` and `availableIPCount` to `IPPoolStatus` so users can see pool usage directly on the object, including as `kubectl get ippool` columns.

`capacity` is the sum of `GetPoolSize()` across `spec.pools`. `availableIPCount` is `capacity` minus the number of addresses currently in use (metal3 and CAPI allocations, including active preallocations). The counts are recalculated on every reconcile, including during deletion as addresses are freed. When the combined size overflows int64, `capacity` is reported as the int64 maximum, `availableIPCount` is omitted, and `capacityOverflow` is set to true.

Fixes #161

**Checklist:**


api/v1alpha1/ippool_types.go
```diff
@@ -108,6 +108,17 @@ type IPPoolStatus struct {
 
 	// Allocations contains the map of objects and IP addresses they have
 	Allocations map[string]IPAddressStr `json:"indexes,omitempty"`
+
+	// TotalIPCount is the total number of allocatable IP addresses summed
+	// across all the pools defined in the spec.
+	// +optional
+	TotalIPCount *int64 `json:"totalIPCount,omitempty"`
```
lentzi90: Maybe "capacity" would make this clearer? With "total" we have to read the description to understand if it is the number of allocated or allocatable or available. I would suggest something like this instead:


```suggestion
	// Capacity is the total number of IP addresses that
	// all the pools defined in the spec can hold.
	// +optional
	Capacity *int64 `json:"capacity,omitempty"`
```

lentzi90 (commented):
Looks quite good!
I suggest we use "capacity" instead of "total IP count".
It would also be nice to have a few more test cases:
- Huge pool, using IPv6
- Zero in-use addresses
- Mixed IPv4 and IPv6

Perhaps the biggest issue here is actually the `int64` and the large numbers we may end up with. I think we need to at least add tests to verify the behavior. With IPv6 we can easily go out o bounds so that we get negative values or strange behavior.

api/v1alpha1/ippool_types.go
migiFi: Kept it as int64, the guard removes the overflow/negative behavior you mentioned and added tests for it. 

The one thing is that for pools too large to fit int64 capacity/availableIPCount are omitted rather than shown. That seems reasonable to me since I dont think we want to display such count? but let me know if you rather cap or represent it differently.

lentzi90: Sounds good! Could you add some note about this also in the doc string?

ipam/ippool_manager.go
```diff
@@ -242,9 +243,52 @@ func (m *IPPoolManager) UpdateAddresses(ctx context.Context) (int, error) {
...
+				m.IPPool.Status.AvailableIPCount = nil
+				m.updateStatusTimestamp()
+			}
+			return fmt.Errorf("failed to get pool size: %w", err)
+		}
+		// GetPoolSize bounds each pool to int64, guard the sum so it can't wrap to a negative.
+		if total > math.MaxInt64-int64(size) {
+			if m.IPPool.Status.Capacity != nil || m.IPPool.Status.AvailableIPCount != nil {
+				m.IPPool.Status.Capacity = nil
+				m.IPPool.Status.AvailableIPCount = nil
```
Rozzii: Availabe should go to 0 not nil.
Capacity should go to maxInt.
We need a new status field called "overflow"  "Pool overflow" or something like that and that should be set to true.

With these 3 values set it is hard to miss that the pool has overflown. It is an other question how to recocile an overflown pool.

All claims should be denied if AvailableIPCount is 0 or below (can't be below but in principle...) and overflow can be cleared only if AvailableIPCount goues above 0.  

AvailableIPCount = 0 does not equal overflow though. But every time AvailableIPCount is above 0 overflow status has to be checked and "cleared".

Rozzii (commented):
The arithmetic overflow scenario has been handled but from UX/design perspecitve I think there are a few adjustments that have to be made to handle the situation and the recovery better.

ipam/ippool_manager.go
Rozzii: In general I would recommend avoiding "nil" status values for integer and boolean status fields. For spec integers could make sense with a nil value if setting the spec to any Int would trigger an operation. I would say the same does not apply for statuses. A status Int value conveys the "quantity /  cardinality" of something  thus  if the "something" doesn't exist or then the quantity or cardinality is not unknown but 0.

It could be argued that "nil" would make sense if there was an error during the retrieval or calculation of said quantity but it is not the case here.

ipam/ippool_manager.go
```diff
@@ -242,9 +243,52 @@ func (m *IPPoolManager) UpdateAddresses(ctx context.Context) (int, error) {
...
+// is the total minus the number of the addresses currently in use (inUse).
+func (m *IPPoolManager) updateAddressCounts(inUse int) error {
+	var total int64
+	for _, pool := range m.IPPool.Spec.Pools {
+		size, err := ipamv1.GetPoolSize(pool)
+		if err != nil {
+			// Clear any previously published counts so we don't leave stale
+			// values when the pool spec can no longer be sized.
+			if m.IPPool.Status.Capacity != nil || m.IPPool.Status.AvailableIPCount != nil {
+				m.IPPool.Status.Capacity = nil
```
Rozzii: Here I would say nil makes sense as the Pool size returned an error so the controller can't even calculate capacity and available  count.

Here the situation is more complicated,  the error content would matter because if the error was induced as a result an overflow inside GetPoolSize then we should set capacity to max int and available count to 0 and set Overflow to true, but if the getPoolSize was simply unable to retrieve the values than the "nil" values do apply.

ipam/ippool_manager.go
migiFi: I can can add an overflow flag, sounds good.

A few notes just so we are in the same page:

The counts are only for display right now, allocation scans the pool directly and doesnt look at AvailableIPCount. Setting it to 0 on overflow is a bit confusing since thats also what a full pool shows, but the overflow flag would tell them apart.

With the flag we could just leave available out and cap capacity and MaxInt64.

Denying claims when available <= 0 would also block allocations on an overflowed pool, even though its a big pool that still works fine. 

@lentzi90 was ok with just leaving the count out on overflow, of course I can make the necessary changes that we agree on.

ipam/ippool_manager.go
migiFi: Agree on this, the issue is that GetPoolSize returns a plain error, so I would have it return a separate error we can check for. 

The sum guard below that is the same case. Still this depends on what we settle on above, but the split makes sense either way.

ipam/ippool_manager.go
Rozzii: I will go with the assumption that by full pool you mean "exhausted/empty" pool, a pool that is full in a sense that it has not allocated any of its IPs should show the actual  AvailableIPCount, if that is not the case then it has to be fixed .

An empty/exhausted pool and an overflown pool is the same from availability perspective IMO, meaning there is no availability. IMO if we handle the fault scenarios correctly the practical difference between an "exhausted" and an "owerflown" pool is just the overflown flag that means "there was at least one attempted allocation after the the pool was fully exhausted" , it is a subtle way of showing that the pool is incorrectly sized for the current use case so the user should pay attention.

In practice no actual "overflow" will happen as the controller would block the attempt , but  the controller will raise the flag signaling that there was an attempt to calim more addresses than available.

migiFi: agreed a countable pool (even fully exhausted one) should show its real `AvailableIPCount`, so `0` when exhausted, that works as is.

The flag I added covers a different case: a pool too large to count (e.g. an IPv6/64). That one's basically empty, not exhausted, so `available=0` would mislead, I cap `capacity` at MaxInt64 and leave available unset. Name it `capacityOverflow` to keep that clear. 

Your case: a claim arriving after the pool is exhausted, is a separate, useful sizing signal. I would add it as its own flag on the allocation path in a follow-up PR rather than mixing it into the capacity counts? what do you think?. Want me to do that as a follow-up or add it here?

api/v1alpha1/ippool_types.go
```diff
@@ -108,6 +108,28 @@ type IPPoolStatus struct {
...
 	// Allocations contains the map of objects and IP addresses they have
 	Allocations map[string]IPAddressStr `json:"indexes,omitempty"`
+
+	// Capacity is the total number of IP addresses that all the pools defined
+	// in the spec can hold. If the combined size of the pools is too large to
+	// represent as an int64 (for example a very large IPv6 subnet such as a
+	// /64 or larger), Capacity is reported as the int64 maximum and Overflow is
+	// set to true. If the pools cannot be sized at all, Capacity is omitted.
+	// +optional
+	Capacity *int64 `json:"capacity,omitempty"`
```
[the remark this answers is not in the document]
migiFi: Kept it as capacity (matching docs and the existing convention) and aligned PR title/description accordingly.

ipam/ippool_manager.go
```diff
@@ -242,9 +243,82 @@ func (m *IPPoolManager) UpdateAddresses(ctx context.Context) (int, error) {
...
+
+	available := total - int64(inUse)
+	if available < 0 {
+		available = 0
+	}
+
+	m.IPPool.Status.CapacityOverflow = false
+	m.IPPool.Status.Capacity = &total
+	m.IPPool.Status.AvailableIPCount = &available
+	return nil
```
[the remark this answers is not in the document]
migiFi: fixed, the normal path now updates status and `LastUpdate` only when the counts actually change.

ipam/ippool_manager.go
```diff
@@ -242,9 +243,82 @@ func (m *IPPoolManager) UpdateAddresses(ctx context.Context) (int, error) {
...
+func (m *IPPoolManager) setCapacityOverflow() {
+	maxCapacity := int64(math.MaxInt64)
+	if m.IPPool.Status.CapacityOverflow &&
+		m.IPPool.Status.Capacity != nil && *m.IPPool.Status.Capacity == maxCapacity &&
+		m.IPPool.Status.AvailableIPCount == nil {
+		return
+	}
+	m.IPPool.Status.CapacityOverflow = true
+	m.IPPool.Status.Capacity = &maxCapacity
+	m.IPPool.Status.AvailableIPCount = nil
```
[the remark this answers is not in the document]
migiFi: updated `docs/api.md` to match the agreedd behavior, on overflow, `capacity` is capped at MaxInt64 with `capacityOverflow=true` and `availableIPCount` omitted, rather than omitting capicity.

kashifest (commented):
/lgtm

ipam/ippool_manager.go
Rozzii: Okay I see you made the flag for the case where even the capacity was too big to manage not when the ip claims are trying to take to many IPs out. 
"I cap capacity at MaxInt64 and leave available unset. Name it capacityOverflow to keep that clear." Yes that I support, cap it at max int and raise the flag. 

It might be very well just me but IMO we should have an other name for this than "CapacityOverflow", it could be confusing to understand what is "overflowing" here from the name of the flag.

I would suggest CapacityRestrained or something similar so it would be clearer that we have forced  limit on the capacity.

Rozzii:
/hold
I am still not 100% okay with it but we are almost there it is shaping up nicely.

ipam/ippool_manager.go
migiFi: I went with `CapacityCapped` instead,  I think it makes it clearer that we are intentionally capping the reported number, not that something overflowed numerically.

how it behaves now:
- Normal pool: real `Capacity`, `AvailableIPCount` is total minus what's in use (0 when it's fully used up), and the flag stays false.
- Pool too big to count (like a /64): `Capacity` is capped at `MaxInt64`, we leave `AvailableIPCount` unset, and the flag is true.
- Pool we can't size at all: we clear the counts and surface the error instead, I split this from the overflow case using the `ErrPoolSizeOverflow` sentinel.
 
The other case you mentioned (flagging when a claim comes in after the pool's already exhausted) I'll do as a follow-up so it doesn't get tangled up with the counts here.

PR #1542, opened 2026-08-10
🐛 Build the release-x.y image when cutting a release

### LINKED ISSUE #1027: Release: when a minor release branch is created, the initial image for tag :release-x is not triggered

Make a minor or RC release by creating appropriate release note PR, which leads to release branch creation upon merging.

A container image is built for the tag `v1.10.0` but also `release-1.10` which is used in testing.

Note that container image build job and release image job do same thing, but only the latter matters in case of release cutting/branch creation due "workflow cannot implicitly initiate another workflow" rule.

**Possible solution:**

Image building in release process is bit different than the push triggered, so it likely is that we need add branch name in release workflow's list of tags to push.

/kind bug

migiFi:
Builds the release-x.y image (e.g. release-1.10) as part of cutting a release, so it exists right after branch creation instead of only appearing on the next push to the branch.

Adds a job that builds the image from the release-branch ref, alongside the existing version-tag build. Skipped for alpha/beta.

**NOTE**: only fully verifiable on real release. The branch-image path is already proven, same reusable workflow used for release-* pushes, just called from the release workflow so it isn't skipped by the `GITHUB_TOKEN` trigger rule.

Fixes #1027

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -17,6 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       release_tag: ${{ steps.release-version.outputs.release_version }}
+      release_branch: ${{ steps.release-branch.outputs.release_branch }}
     if: github.repository == 'metal3-io/ip-address-manager'
     steps:
     - name: Checkout code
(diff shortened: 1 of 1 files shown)


migiFi:
I have set `generate-sbom: false` on the release-branch job to avoid conflict, image is still signed and re-attested on the branch push. 

Can extend the reusable workflow to take a unique SBOM name per ref if we would rather keep it on.

kashifest (commented):
/lgtm

Sunnatillo (commented):
/retest
LGTM

PR #1564, opened 2026-08-21
🐛 Report IPAddress name collisions from duplicate namePrefix

### LINKED ISSUE #1437: IPpools with identical NamePrefix fail IPAddress allocation due to name collision
```bash
# Create cluster
kind create cluster
clusterctl init --infrastructure docker --ipam metal3


# Create namespace, IPPool and IPClaim
kubectl create namespace test-ns

kubectl apply -f - <<EOF
kind: IPPool
metadata:
  name: ippool-a
  namespace: test-ns
spec:
  namePrefix: "a-prefix"
  pools:
    - start: "192.168.1.10"
      end: "192.168.1.100"
      subnet: "192.168.1.0/24"
      prefix: 24
      gateway: "192.168.1.1"
  prefix: 24
  gateway: "192.168.1.1"
---
kind: IPClaim
metadata:
  name: a-claim
  namespace: test-ns
spec:
  pool:
    name: ippool-a
    namespace: test-ns
EOF

# Create second pool with same name prefix and ip range. 
# Observe how creating the second claim silently fails
kubectl apply -f - <<EOF
kind: IPPool
metadata:
  name: ippool-b
  namespace: test-ns
spec:
  namePrefix: "a-prefix"       # <-- duplicates IPPool-A
  pools:
    - start: "192.168.1.10"
      end: "192.168.1.10"
      prefix: 24
---
kind: IPClaim
metadata:
  name: b-claim
  namespace: test-ns
spec:
  pool:
    name: ippool-b
    namespace: test-ns
EOF

kubectl describe ippool -n test-ns ippool-b
# Does not show event for IPAddress creation

kubectl describe ipclaim -n test-ns b-claim
# Doesn not show allocated IP

kubectl get ipaddress.ipam.metal3.io -n test-ns
# Shows only one IPAddress
# NAME                    AGE
# a-prefix-192-168-1-10   8m48s
```

I expect there to be no silent failures. 

The issue happens because Kubernetes API prevents creating object  with same kind/name/namespace combos. IPaddress name is created based on IPPool namePrefex and allocated IP. 

**Environment:**

- Cluster-api version: 1.13
- IPAM version: 1.13 
- environment: kind cluster with `clusterctl init --infrastructure docker --ipam metal3`

/kind bug

migiFi:

When two IPPools in the same namespace share a `spec.namePrefix` and their address ranges overlap, they generate identical IPAddress object names. The second claim's IPAddress creation fail with `AlreadyExist`, which `createObject` converts into a transient error. As a result no status was set on the claim, no event was emitted, and the controller requeued indefinitely, a silent failure that is hard to diagnose.

This change makes the failure observable and stops the requeued loop, without changing the naming scheme.

**Changes**:

- On IPAddress create failure, probe the existing object's owner:
  - Owned by a *different* pool/claim -> treat as a terminal collision.
  - Owned by *this* pool/claim (stale-cache race) -> keep the existing transient/requeue behavior.
- Surface the collision on both objects:
  - IPClaim: `Status.ErrorMessage` (Metal3) / `AllocationFailed` condition (CAPI). Setting these also stops the reconcile loop from re-processing the claim.
  - IPPool: a `Warning` `IPAddressNameCollision` event, via a new `EventRecorder` wired through the manager factory.
- Unit test for the metal3 collision, metal3 benign race, CAPI collision, and the IPPool event emission.

No CRD/RBAC changes and no mock regeneration were needed as events RBAC already existed and the `ManagerFactoryInterface` signature is unchaged.

**NOTE**: this does not let the same `namePrefix` pool with overlapping ranges coexist, the collision claim still gets no address, it just fails loudly with an actionable message. Enabling coexistence requires a bigger change to the IPAddress naming scheme and should be discuss in another issue.

Fixes #1437

**Checklist:**

### DIFF
--- a/ipam/ippool_manager.go
+++ b/ipam/ippool_manager.go
@@ -30,6 +30,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/events"
 	"k8s.io/utils/ptr"
 	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
 	capipamv1 "sigs.k8s.io/cluster-api/api/ipam/v1beta2"
@@ -57,9 +58,10 @@ type IPPoolManagerInterface interface {
 
 // IPPoolManager is responsible for performing machine reconciliation.
 type IPPoolManager struct {
-	client client.Client
-	IPPool *ipamv1.IPPool
-	Log    logr.Logger
+	client   client.Client
+	IPPool   *ipamv1.IPPool
+	Log      logr.Logger
+	recorder events.EventRecorder
 }
 
 // NewIPPoolManager returns a new helper for managing a ipPool object.
@@ -755,13 +757,26 @@ func (m *IPPoolManager) createAddress(ctx context.Context,
 		},
 	}
 
-	// Create the IPAddress object. If we get a conflict (that will set
-	// Transient error), then requeue to retrigger the reconciliation with
-	// the new state
+	// Create the IPAddress object. On AlreadyExists, probe the existing
+	// object's owner: a different pool/claim is a terminal namePrefix
+	// collision (status set, no requeue); the same pool/claim is a stale-cache
+	// race and stays transient so we requeue.
 	if err := createObject(ctx, m.client, addressObject); err != nil {
 		var reconcileError ReconcileError
 		if !errors.As(err, &reconcileError) {
+			// Not the AlreadyExists case: surface the real error and requeue.
 			addressClaim.Status.ErrorMessage = ptr.To("Failed to create associated IPAddress object")
+			return addresses, err
+		}
+		// Name is taken. Probe the owner to tell a real collision from a
+		// benign stale-cache race.
+		if collision, poolName, claimName := m.checkNameCollision(ctx, &ipamv1.IPAddress{}, addressName, addressClaim.Name); collision {
+			msg := fmt.Sprintf("IPAddress name collision: %q already exists, owned by IPPool %q / IPClaim %q (likely a duplicate spec.namePrefix)", addressName, poolName, claimName)
+			addressClaim.Status.ErrorMessage = ptr.To(msg)
+			if m.recorder != nil {
+				m.recorder.Eventf(m.IPPool, nil, corev1.EventTypeWarning, "IPAddressNameCollision", "AllocateIPAddress", "%s", msg)
+			}
+			return addresses, WithTerminalError(errors.New(msg))
 		}
 		return addresses, err
 	}
@@ -861,12 +876,14 @@ func (m *IPPoolManager) capiCreateAddress(ctx context.Context,
 		},
 	}
 
-	// Create the IPAddress object. If we get a conflict (that will set
-	// Transient error), then requeue to retrigger the reconciliation with
-	// the new state
+	// Create the IPAddress object. On AlreadyExists, probe the existing
+	// object's owner: a different pool/claim is a terminal namePrefix
+	// collision (status set, no requeue); the same pool/claim is a stale-cache
+	// race and stays transient so we requeue.
 	if err := createObject(ctx, m.client, addressObject); err != nil {
 		var reconcileError ReconcileError
 		if !errors.As(err, &reconcileError) {
+			// Not the AlreadyExists case: surface the real error and requeue.
 			conditions := make([]metav1.Condition, 0, 1)
 			conditions = append(conditions, metav1.Condition{
 				Type:               capipamv1.IPAddressClaimReadyCondition,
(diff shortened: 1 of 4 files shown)


ipam/ippool_manager.go
```diff
@@ -1070,3 +1094,31 @@ func anyErrorInExistingClaim(addressClaim capipamv1.IPAddressClaim) bool {
...
+// isNameCollision returns true if an IPAddress with the given name already
+// exists but is NOT owned by this pool/claim (i.e. a namePrefix collision),
+// along with the owning pool and claim names for the error message.
+func (m *IPPoolManager) isNameCollision(ctx context.Context, name string) (bool, string, string) {
+	existing := &ipamv1.IPAddress{}
+	key := client.ObjectKey{Name: name, Namespace: m.IPPool.Namespace}
+	if err := m.client.Get(ctx, key, existing); err != nil {
+		return false, "", ""
+	}
+	if existing.Spec.Pool.Name == m.IPPool.Name && existing.Spec.Claim.Name != "" {
```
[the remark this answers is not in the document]
migiFi: Updated both helpers to take the claim name and treat the failure as benign only when the existing IPAddress is own by this exact pool and claim, so a same-pool/different-claim address is now correctly classified as a terminal collision instead of looping. 

Added a unit test covering the same-pool, different-claim case.

ipam/manager_factory.go
```diff
@@ -30,15 +31,21 @@ type ManagerFactoryInterface interface {
...
-	client client.Client
+	client   client.Client
+	recorder events.EventRecorder
 }
 
 // NewManagerFactory returns a new factory.
-func NewManagerFactory(client client.Client) ManagerFactory {
-	return ManagerFactory{client: client}
+func NewManagerFactory(client client.Client, recorder events.EventRecorder) ManagerFactory {
+	return ManagerFactory{client: client, recorder: recorder}
```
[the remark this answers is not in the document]
migiFi: This function is only used here so the change wouldnt actually break anything. But I have made it compatible anyway, the original `NewManagerFactory(client)` call still works, and the event recorder is now optional.

tuminoid (commented):
/lgtm
/cc @adilGhaffarDev @lentzi90

ipam/ippool_manager.go
```diff
@@ -759,6 +761,14 @@ func (m *IPPoolManager) createAddress(ctx context.Context,
 	// Transient error), then requeue to retrigger the reconciliation with
 	// the new state
```
kashifest: This comment is not valid anymore I guess? We are setting terminal error now?

ipam/ippool_manager.go
```diff
@@ -1070,3 +1094,33 @@ func anyErrorInExistingClaim(addressClaim capipamv1.IPAddressClaim) bool {
...
+		return false, "", ""
+	}
+	// Benign only if this exact pool+claim already owns the address (stale-cache race).
+	if existing.Spec.Pool.Name == m.IPPool.Name && existing.Spec.Claim.Name == claimName {
+		return false, "", ""
+	}
+	return true, existing.Spec.Pool.Name, existing.Spec.Claim.Name
+}
+
+// capiIsNameCollision is the CAPI-typed equivalent of isNameCollision.
```
kashifest: why do we need two different functions for almost identical task?  cant we just if and else on the `existing` initialization?

ipam/ippool_manager.go
```diff
@@ -865,6 +875,20 @@ func (m *IPPoolManager) capiCreateAddress(ctx context.Context,
 	// Transient error), then requeue to retrigger the reconciliation with
```
kashifest: Same here

kashifest (commented):
Few comments inline

ipam/ippool_manager.go
migiFi: fixed, reworded the comment in both create paths to cover the terminal-collision vs stale-cache-race split.

ipam/ippool_manager.go
migiFi: No special reason, I have now merged them into one `checkNameCollision` with a type switch. callers pass the empty typed object. Only tradeoff is it takes `client.Object` so the type check moves the runtime (unknown type -> no collision).

ipam/ippool_manager.go
```diff
@@ -755,10 +757,19 @@ func (m *IPPoolManager) createAddress(ctx context.Context,
...
 	if err := createObject(ctx, m.client, addressObject); err != nil {
+		if collision, poolName, claimName := m.checkNameCollision(ctx, &ipamv1.IPAddress{}, addressName, addressClaim.Name); collision {
+			msg := fmt.Sprintf("IPAddress name collision: %q already exists, owned by IPPool %q / IPClaim %q (likely a duplicate spec.namePrefix)", addressName, poolName, claimName)
+			addressClaim.Status.ErrorMessage = ptr.To(msg)
+			if m.recorder != nil {
+				m.recorder.Eventf(m.IPPool, nil, corev1.EventTypeWarning, "IPAddressNameCollision", "AllocateIPAddress", "%s", msg)
+			}
+			return addresses, WithTerminalError(errors.New(msg))
+		}
 		var reconcileError ReconcileError
```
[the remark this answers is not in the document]
migiFi: Fixed, probe now only runs on the transient AlreadyExist error so non-transient failures arent masked. Added a test.

ipam/ippool_manager.go
```diff
@@ -1070,3 +1096,35 @@ func anyErrorInExistingClaim(addressClaim capipamv1.IPAddressClaim) bool {
...
+// Pass an empty typed object (&ipamv1.IPAddress{} or &capipamv1.IPAddress{})
+// so the same Get + comparison logic serves both APIs; only the owner-ref
+// field paths differ between the two IPAddress types.
+func (m *IPPoolManager) checkNameCollision(ctx context.Context, existing client.Object, name, claimName string) (bool, string, string) {
+	key := client.ObjectKey{Name: name, Namespace: m.IPPool.Namespace}
+	if err := m.client.Get(ctx, key, existing); err != nil {
+		return false, "", ""
+	}
+
+	var ownerPool, ownerClaim string
```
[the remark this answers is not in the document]
migiFi: Fixed, `checkNameCollision` treats a terminating object as no-collision so it stays transient. Test added.

PR #1578, opened 2026-09-04
:bug: Fix set-manifest-image for IPv6 addresses

nuhakala:

Adds quotation marks around the image in set-manifest-image makefile target. The quotation marks were removed in #1535 because no-one remembered what they were needed for. If the image contains registry IP address and that IP address is IPv6 (hence contains brackets), the yaml formatting will break.

**Checklist:**

- [x] E2E tests have been added, if necessary.


kashifest (commented):
/copilot-review

kashifest (commented):
/approve

Rozzii (commented):
/lgtm

Rozzii:
@nuhakala wirte a comment in the makefile please explaining why the chars shouldn't be removed

nuhakala:
> @nuhakala wirte a comment in the makefile please explaining why the chars shouldn't be removed

Added

peppi-lotta:
Please add the comment to both of the targets (just in case the one with the comment ever gets removed)

nuhakala:
> Please add the comment to both of the targets (just in case the one with the comment ever gets removed)

Added, @peppi-lotta

